PR #3, opened 2014-12-17
Detect Coffee version and use correct compiler switch

p0deje:
For CoffeeScript >= 1.7.0 we should pass `--compilers coffee:coffee-script/register`. This pull request checks version the first time it's run and use correct compiler switch.

I haven't really tested CoffeeScript < 1.7.0 so if you could take a look that would be great!

### DIFF
--- a/autoload/test/mocha.vim
+++ b/autoload/test/mocha.vim
@@ -18,7 +18,7 @@ function! test#mocha#build_args(args) abort
   let args = a:args
 
   if !empty(glob('**/*coffee'))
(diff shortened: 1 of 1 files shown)


janko-m:
I [removed](https://github.com/janko-m/vim-test/commit/6555be646dd4d1615c7ff16179b8ff82b12490a4) support for CoffeeScript because of issues in described in https://github.com/janko-m/vim-test/issues/52. For mocha you can just set the compiler in test/mocha.opts, which I think should be done anyway.

PR #16, opened 2015-03-06
Update gotest.vim, fix -run flag, run from any file

jgeiger:
- Fix go test -run. If we run a file, only use the file name. If we run
  a targeted test, don't include the file name.

### DIFF
--- a/autoload/test/go/gotest.vim
+++ b/autoload/test/go/gotest.vim
@@ -6,9 +6,9 @@ function! test#go#gotest#build_position(type, position) abort
   if a:type == 'nearest'
     let name = s:nearest_test(a:position)
     if !empty(name) | let name = '-run '.shellescape(name, 1) | endif
-    return [a:position['file'], name]
+    return [name]
   elseif a:type == 'file'
-    return [a:position['file']]
+    return ['-run', a:position['file']]
   else
     return []
   endif


janko-m:
First of all, thank you for the pull request :)

Could you elaborate more, why do you want to treat all .go files as test files? I like that currently if you're editing a .go file, then `:TestNearest`, `:TestFile` and `:TestSuite` will automatically execute last touched `*_test.go` file.

If you exclude the filename from the nearest test, wouldn't it be possible that another file which has a test function with the same name also gets run? Does `-run` regexp also match the file names? Why is using `-run` for file tests better?

jgeiger:
First, I reverted the change to the file types. I see what you did, and it works with a bit of thought and running the test file first. I'm used to Ruby/Rails where I can run the spec for a file while still inside that file.

It seems like go is a bit picky about how you run the tests.

    go test
runs the entire suite 

    go test -run
accepts a regex and runs test that match, which works for focused tests

It seems like there isn't a go compatible way to run all the tests in single file without doing some magic about including packages.

Either way, the way the code exists without this patch will not correctly run a targeted test. (or at least I cannot get it to work without my changes)

jgeiger:
I need to further check something later today to see why my code was running the single file test correctly. I've been informed that it shouldn't be working and might just be a fluke.

janko-m:
> I see what you did, and it works with a bit of thought and running the test file first. I'm used to Ruby/Rails where I can run the spec for a file while still inside that file.

I originally meant to implement that when you run test while in an application file, that it switches to the corresponding test file. However, for some reason I found it really difficult to implement, because I couldn't determine what the user would want (if you run some test, and then you switch to an unrelated application file and run tests, should I run the tests for that new file or the last test?). Because of these hard questions I rather decided not to include it in test.vim, because then users will what to expect and aren't suprised. In Rails projects people can simply run `:A` to switch to the test file, and I decided that it is satisfactory enough.

PR related, you're the one who knows Go, so if you say it works, it works :wink:. I was just asking to check. It also works for me when I try it. Thank you again for the PR!

PR #17, opened 2015-03-07
"Fix" go testing options since go doesn't test single files

jgeiger:
- 'file' option will test all the *_test.go files in the current directory by running 'go test'.
  Go doesn't have a concept of running all tests in a single file without some tricky file inclusion.

- 'suite' option will run all the tests in the current directory and all subdirectories by running 'go test ./...'.

### DIFF
--- a/autoload/test/go/gotest.vim
+++ b/autoload/test/go/gotest.vim
@@ -8,9 +8,9 @@ function! test#go#gotest#build_position(type, position) abort
     if !empty(name) | let name = '-run '.shellescape(name, 1) | endif
     return [name]
   elseif a:type == 'file'
-    return ['-run', a:position['file']]
-  else
     return []
+  else
+    return ['./...']
   endif
 endfunction
 


jgeiger:
After checking my code that I thought was working for a single file, I realized I was getting a false pass message.

I've updated the code to do something that might make more sense.

If there are any other Go devs out there, I'd be happy to have them weigh in.

janko-m:
Hmm, when I was doing my local tests for running tests for a file, I commented out the only test in that file, and then I was getting a "no tests were run" message from Go. So for me that was kind of a proof that things are working correclty. How did you conclude that you were getting a false pass?

If `go test` doesn't run the whole test suite, what does it run then?

janko-m:
@mislav, can you help us? We are figuring out how to run Go tests on different granularities. We currently have

| Type | Command | Comments |
| ------ | ---------- | ------- |
| Single test | `go test -run TestSomething` | My concern is that this will run all tests that are named the same. However, passing in a filename [assumes packages](http://stackoverflow.com/questions/16935965/how-to-run-test-cases-in-a-specified-file#answer-16936314). |
| File test | `go test -run something_test.go` | `go test something_test.go` also assumes packages. Is this the correct way, then? |
| Suite test | `go test` | Does this run the whole test suite? @jgeiger wishes to change it to `go test ./...`.

mislav:
@jgeiger's original ideas are :+1:

janko-m:
Thank, bro. Ok, great, merging in :)

PR #21, opened 2015-03-26
Add neovim terminal strategy

michaldarda:
Hey!

I really enjoy your plugin in my day to day work, I would love to contribute.

Neovim awesome build-in terminal was just merged to master (https://github.com/neovim/neovim/pull/2076#event-265371619)

I quickly hacked a new strategy for running tests inside it's term using your plugin, it works for me, but it's surely not perfect, so any suggestions (for example for addings some tests) will be welcomed :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,6 +69,15 @@ Runs test commands with `:!`, which switches your Vim to the Terminal.
 let g:test#strategy = 'basic'
 ```
 
+### Neovim terminal
+
+Runs test commands with `:terminal`, which is awesome built-in neovim terminal.
+Requires neovim with terminal patch (https://github.com/neovim/neovim/pull/2076)
+
+```vim
+let g:test#strategy = 'neovim'
+```
+
 ### Dispatch.vim
 
 Runs test commands with `:Dispatch`. Requires the
--- a/autoload/test/strategy.vim
+++ b/autoload/test/strategy.vim
@@ -31,7 +31,7 @@ function! test#strategy#iterm(cmd) abort
 endfunction
 
 function! test#strategy#neovim(cmd) abort
-  execute 'terminal'.s:pretty_command(a:cmd)
+  execute 'terminal '.s:pretty_command(a:cmd)
 endfunction
 
 function! s:execute_script(name, cmd) abort
--- a/autoload/test/strategy.vim
+++ b/autoload/test/strategy.vim
@@ -30,6 +30,10 @@ function! test#strategy#iterm(cmd) abort
   call s:execute_script('osx_iterm', s:pretty_command(a:cmd))
 endfunction
 
+function! test#strategy#neovim(cmd) abort
+  execute 'terminal '.a:cmd
+endfunction
+
 function! s:execute_script(name, cmd) abort
   let script_path = g:test#plugin_path . '/bin/' . a:name
   let cmd = join([script_path, shellescape(a:cmd)])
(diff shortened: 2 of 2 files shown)


janko-m:
That's an awesome idea, thank you!

Could you just add it to the readme (I think right after the "basic" strategy would be good, since it's similar). And would be cool if you could also add it to the `doc/test.txt`.

I can't seem start Neovim compiled from the latest commit, there are some errors loading the Python host, so I can't try it out. How does `:terminal` actually work? Is it similar to `:!`? Does the Terminal perist between tests? If yes, maybe we should also use `s:pretty_command()`, which outputs the command and clears the screen?

janko-m:
Ok, even though Neovim is erroring, but I can still run the `:terminal`. It seems to work in the same way as `:!`, so just wrap the `a:cmd` in `s:pretty_command()`, like it is on the basic strategy (but ignore that conditional).

michaldarda:
@janko-m updated readme

I also tried to use pretty_command and got this error when running test

Not an editor command: terminalclear; echo -e 'bundle exec rspec spec/calls/get_objects/offers_json_spec.rb'; bundle exec rspec spec/calls/get_objects/offers_json_spec.rb

janko-m:
It should be `'terminal '.pretty_command(a:cmd)` (you forgot the space). Don't forget `doc/test.txt` as well ;). And at the end I would like that you squash everything into a single commit :)

michaldarda:
Thanks ;)

I will squash commits.

But now I've got
E116: Invalid arguments for function termopen :(

janko-m:
Ok, was reading Neovim's PR a bit, and I have different idea. Instead of `:terminal` try `:call termopen(s:pretty_command(:a:cmd))`.

michaldarda:
@janko-m 

It shows error
ml_get: invalid lnum: 31

And then shows spec output

janko-m:
I made a typo, it's `a:cmd`, not `:a:cmd`, you probably already corrected it. That's strange, if I just do a direct

```vim
:call termopen("clear; echo -e 'bundle exec rspec spec/calls/get_objects/offers_json_spec.rb'; bundle exec rspec spec/calls/get_objects/offers_json_spec.rb")
```

It properly executes the command. Does it work for you too? As I understood, you get the error, but the commands still executes properly?

michaldarda:
@janko-m yes, I've corrected it :)

Command that you have provided executed without any errors. The only only strange thing is that it show press any key to close, but in fact I must explicitly enter :q to close vim pane with test output

michaldarda:
@janko-m I must admit I like the original version
```execute 'terminal '.a:cmd```

because I could insert pry inside my rspec tests and interactivily debug inside vim, as I understand correctly using pretty_command this is not possible?

janko-m:
I think you should still be able to `binding.pry` with pretty_command. I raised an issue on Neovim to help me with starting it up, so that I can finally test it myself.

janko-m:
You're right, I didn't know that `:call termopen()` is different from `:terminal`. Switch it back to `execute 'terminal '.a:cmd`, and I'm merging. If we can't get the command to be echoed, it's fine, not a big deal anyways.

janko-m:
Damn, the thing is that even if we don't use pretty_command, `:terminal` will still trip on test commands which have command-line options in it (anything with quotes), which will happen very often. I really appreciate your work on this PR, but I would like to go back to this when `:terminal` gets more stable and documented.

janko-m:
The docs for `:terminal` [emerged](http://neovim.org/doc/user/various.html#:terminal), so I was able to add the neovim strategy :) (and authored you on the commit).

PR #22, opened 2015-03-29
Add VimTmuxRunner strategy

jackfranklin:
Just swapped to using VTR for my Vim + Tmux integration and missed having support for it in this!


janko-m:
Awesome, thank you.

PR #24, opened 2015-04-10
Add neoterm strategy

wordofchristian:
Neoterm provides better customization options for how terminals are spawned in
neovim. It behaves more like Dispatch in that it can send the command to a new
split or an existing split if there is one already open. Neoterm has some built in test 
commands but they aren't as good as the ones provided by vim-test. This strategy 
allows you to get the best of both worlds.


janko-m:
Cool, thanks! :heart:

There is just one typo in the docs, "deisplay", if you could just correct it :)

janko-m:
I would normally ask you to squash the commits into a single commit, but to save us trouble, I want to try Tim Pope's style of merging PRs.

https://github.com/janko-m/vim-test/commit/45074b87b76aff2c35db3ff03c2c8eaca1660b37

Issue #29, opened 2015-04-28
Name of class incorrect when doing a TestNearest in python, contains inheritance info


yannack:
I am trying to run TestNearest on a python nosetest. Here is the command which is being transferred to the neighbor tmux pane:
```
nosetests --doctest-tests test/scanner_manager/test_ballot_processor.py:TestBallotProcessor\(unittest\.TestCase\).test_unmatch_db_ballot_2
```
As you can see, it is considering the class declaration as a subclass of unittest.TestCase to be part of the name of the class, which is incorrect.
TestBallotProcessor\\(unittest\\.TestCase\\) --> TestBallotProcessor

janko-m:
Thanks for reporting. Could you check if this fix works?

yannack:
Seems good to me !

PR #31, opened 2015-05-04
allow underscores and numbers in python class names

nixon:
The fix for issue #29 will still fail when the class name contains
underscores or numbers, both of which are valid in python.  I noticed
that the regex also allows dots, which are not valid in a python class
name ( http://stackoverflow.com/a/10120327 )

### DIFF
--- a/autoload/test/python.vim
+++ b/autoload/test/python.vim
@@ -1,4 +1,4 @@
 let test#python#levels = [
   \ '\v^\s*def (test_\w+)',
-  \ '\v^\s*class ([A-Za-z.]+)',
+  \ '\v^\s*class (\w+)',
 \]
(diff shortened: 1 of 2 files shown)


janko-m:
Thanks for the fix. Hmm, I though that a "dot" is used when nesting classes. I though that `Foo.Bar` is a `Bar` class nested under the `Foo` class, so that it should be given to `nosetests` as a whole (not just `Bar`). I'm not a Python programmer, so you know better than me, I just want to check.

nixon:
That's true for import statements and places like that, but in a class name, python will throw a SyntaxError exception if there is a dot in a class name.

```
$ cat <<EOF > c.py
class Foo.Bar(object):
    pass
EOF

$ python c.py
  File "c.py", line 1
    class Foo.Bar(object):
             ^
SyntaxError: invalid syntax
```

janko-m:
Great, that's all I needed to know. Thanks again!

Issue #27, opened 2015-04-23
Not a test file


nihlaeth:
I've been trying to get vim-test v1.2.0 working with python and nosetests under linux (Slackware 14.1), but all I get is "Not a test file" when running any of the :Test* commands, regardless of the current buffer.

Running :Nose directly works fine.

janko-m:
Does the filename of your test file conform to the "test_*.py" format (e.g. test_foo.py)? Can you give me one test filename which test.vim doesn't recognize as a test file? When your current buffer is on the test file, what output does the following Vim command give you:

```vim
:echo fnamemodify(expand('%'), ':t') =~# '^test_.*\.py$'
```

nihlaeth:
Well, no, as nosetests recognizes test.* as tests files too, I assumed vim.test would as well (finding no documentation to cobradict that). I'll just modify the regex then, thanks!

janko-m:
You know nosetests better than me, I thought that it supported only `test_*.py` files. If it supports other filenames as well, I want to support it to :)

Can you point me to some docs about this, if you know of any? What kind of filenames are `test.*`, is there something after the dot that isn't `.py`?

nihlaeth:
According "man nosetests", the default matching pattern is (?:^|[b_.-])[Tt]est)

That doesn't mention the ending/file extension because it probably does not look at files, but at modules that can be imported from within python. You're probably safe if you go for (?:^|[b_.-])[Tt]est).*\.py as match.

nihlaeth:
As an afterthought, nose will try to find tests in any file you feed it. The matching is only used if you don't specify any. Is there a reason why you want to do your own matching here?

Personally, I have some tests I do NOT want executed when running the whole suite, as they do extensive profiling, or take a long time for testing performance and stuff. I usually just give those a name that does not get through nose's matching, but I might want to run them with vim.test. Nose does some matching in function/class/method naming as well, so you're pretty safe, throwing any .py file at it.

janko-m:
The reason why I do pattern matching in general is because test.vim works in a way that it tries to recognize which file is a test, and which is application code. And when you're in an application file and you run your tests, test.vim runs the tests from the last test file. This very useful when you're trying to make your tests pass, that you don't have to always switch to the test file to run tests.

For that reason I need to be able to recognize between the two.

janko-m:
I pushed now a fix with a looser pattern for Nose's test files (I took your recommendation slightly simplified).

[...]

PR #32, opened 2015-05-04
allow "test" to be anywhere in a filename

nixon:
by default, nose looks for a file with "test" at the beginning of the
name, or anywhere in the filename after a word boundary like underscore
or dot.

http://nose.readthedocs.org/en/latest/api/config.html

### DIFF
--- a/autoload/test/python/nose.vim
+++ b/autoload/test/python/nose.vim
@@ -1,5 +1,5 @@
 function! test#python#nose#test_file(file) abort
-  if fnamemodify(a:file, ':t') =~# '\v[Tt]est.*\.py$'
+  if fnamemodify(a:file, ':t') =~# '\v(^|[\b_\.-])[Tt]est.*\.py$'
     if exists('g:test#python#runner')
       return g:test#python#runner == 'nose'
     else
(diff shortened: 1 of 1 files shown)


janko-m:
Would it maybe work just to copy-paste that regexp? In #27 it was said that `(?:^|[b_.-])[Tt]est)` is Nose's regex, but I see here that it's a bit more complex. I decided to use a simpler (less permissive) regex, but maybe I shouldn't have tried to be a smartass and I should have just used Nose's regex directly.

nixon:
I'm not sure what the equivalent vim regex syntax would be.  The nose regex actually has a `\b` to match word boundaries and also includes the operating system directory separator (`/` for unix, `\` for windows).  I decided to make my pull request just look for the word "test" because it seemed simpler.

I've always named my test files as `example_test.py` so that bash's tab complete is easier.  Any regex that works with that is ok with me.  :)

janko-m:
Vim's regexes should work like POSIX regexes if you add `\v` at the beginning. Could you maybe update the regex to Nose's one? I think then we can be certain.

janko-m:
Thanks!

PR #35, opened 2015-05-16
Add MoonScript support for Busted

kidanger:


janko-m:
Thanks for the PR. I edited your commit a bit, removing the entry in the README, because I prefer to consider MoonScript as Lua, for the same reason I prefer considering CoffeeScript as JavaScript.

https://github.com/janko-m/vim-test/commit/c15369f945aab3bfa2209a10a1bce4c671e1d1d9

PR #37, opened 2015-06-04
Update mocha matching regex

MichaelJosephKramer:
This update allows the mocha test file regex to match a nested test
directory.


janko-m:
Thanks!

PR #42, opened 2015-07-28
PHPUnit support

codeinabox:
I am trying to create a pull request to add PHPUnit support but at the moment I am having trouble getting my tests to pass as `Expect exists('g:test#last_command') == 1` is failing. Is there something else I need to do to hook it in?

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,7 @@ Currently the following testing frameworks are supported:
 | **Shell**      | Bats                                  |
 | **VimScript**  | VSpec, Vader.vim                      |
 | **Lua**        | Busted                                |
+| **PHP**        | PHPUnit                               |
 
 ## Idea
 
--- a/autoload/test/php.vim
+++ b/autoload/test/php.vim
@@ -1,0 +1,4 @@
+let test#php#levels = [
+  \ '\v^\s*function (test\w+)',
+  \ '\v^\s*class (\w+)',
+\]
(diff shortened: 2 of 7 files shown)


janko-m:
In `test#php#phpunit#test_file` you used the `'test\.php$'` regex, so vim-test doesn't recognize `NormalTest.php` as a PHPUnit test file (because of the capital T). If you change the regexp to `'\v(t|T)est\.php$'`, it should work.

codeinabox:
Even after that fix still having issues with the tests failing 

> -------- Testing a Vim plugin
spec/phpunit_spec.vim .. 1/?
not ok 1 - PHPUnit runs file tests instead of nearest tests
 Expected exists('g:test#last_command') == 1
       Actual value: 0
     Expected value: 1
not ok 2 - PHPUnit runs file tests
 Expected exists('g:test#last_command') == 1
       Actual value: 0
     Expected value: 1
not ok 3 - PHPUnit runs test suites
 Expected exists('g:test#last_command') == 1
       Actual value: 0
     Expected value: 1
spec/phpunit_spec.vim .. Failed 3/4 subtests

janko-m:
Sorry for the slow responses :). in `plugin/test.vim` you also need to add PHPUnit to the list in the `g:test#runners` variable. Adding the following line should do it:

```vim
  \ 'PHP':        ['PHPUnit'],
```

Also, after making it work, you can add it to the list of frameworks in the README.

janko-m:
Excellent! Ready for merging?

codeinabox:
I think so, I've tested it out on a project and it's working fine for me.

janko-m:
Awesome, thank you so much for contributing! :heart:

PR #46, opened 2015-07-29
Behat WIP

codeinabox:
I've got the Behat tests working for everything except `TestNearest`. If you're cursor is in the middle of the scenario it needs to pick the line that "Scenario:" is on. How do I do this with `let test#php#levels`?

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ Currently the following testing frameworks are supported:
 | **Shell**      | Bats                                  |
 | **VimScript**  | VSpec, Vader.vim                      |
 | **Lua**        | Busted                                |
-| **PHP**        | PHPUnit                               |
+| **PHP**        | PHPUnit, Behat                        |
 
 ## Idea
 
--- a/autoload/test/php.vim
+++ b/autoload/test/php.vim
@@ -1,4 +1,5 @@
 let test#php#levels = [
   \ '\v^\s*function (test\w+)',
   \ '\v^\s*class (\w+)',
+  \ '\v^\s*Scenario:\s*',
 \]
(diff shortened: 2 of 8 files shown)


janko-m:
Aha, Behat doesn't support line numbers like Cucumber does.

There is `test#base#nearest_test(position, levels)` to which you pass the "position" dictionary and an array of levels (regexes), and it returns a "namespace" array it found, which you can then join in whatever way the framework requires. The regex is important, the 1st matchgroup will be the thing it is returned in the "namespace" array. So, basically, you need to put the pattern for the string you want inside of parentheses, in your case `Scenario: (.*)`, because you want the string that comes after "Scenario".

I hope I explained it well, I know I should improve the contributing documentation for this part, ask me if you get stuck. I'm really impressed how you figured all of the other things out.

janko-m:
Btw, I think you don't need to put the `Scenario: (.*)` regex in `test#php#levels`, I think you can just inline it in `autoload/test/php/behat.vim`:

```vim
namespace = test#base#nearest_test(position, ['Scenario: (.*)'])
namespace[0] " This is your Scenario string
```

janko-m:
Actually, hold off on this for a bit, I need to do some refactoring of this part, it's messy now. I will then merge this and do the changes for nearest test.

janko-m:
https://github.com/janko-m/vim-test/commit/1392bbee2aab3adc7ed0bfc1f833bb36c6304b55

Thank you for the PHP contributions! :heart:

codeinabox:
I can test this against some real Behat features tomorrow. Do you want me to squash all the commits when I'm done?



—

janko-m:
I already cherry-picked your commit and updated the nearest test part (your changes are in the master now, I linked to your commit). Yeah, if you find any bugs against real Behat features, it would be great if you submit patches.

PR #47, opened 2015-07-30
Look for phpunit in standard Composer ./vendor/bin and ./bin directories

codeinabox:
This updates the phpunit runner to look for phpunit in standard Composer ./vendor/bin and ./bin directories


janko-m:
Thanks!

Issue #28, opened 2015-04-25
Defining your own custom strategies


janko-m:
From @rschmukler:

> @janko-m how would someone create a custom strategy for vim-test? Specifically, when using something like Vundle which uses version control to manage plugins (so you don't want to edit the strategies.vim file)? How do we create a custom strategy given that it appears to be using test#strategy#\<name\> namespacing?

jalcine:
This is an compelling idea. I'd like to see if I could rig my CMake plugin to test compatible projects.

janko-m:
Solved by #51.

PR #51, opened 2015-08-08
Add support for custom strategies

zenbro:
First, thanks for this lifechanging plugin. :smiley: 

This is a simple support for custom strategies (#28) via dict and funcref.
Docs and readme are also updated.


janko-m:
Wow, this is so awesome! I was delaying solving this issue for a long time, and you just went ahead and solved it. Thank you! :heart:

PR #54, opened 2015-09-02
Added support for Django tests.

benjohnson:
YAY! Django tests!

I've been using this fantastic plugin for a year with Django and figured it was about time I wrote some tests and PR'd the thing :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ Currently the following testing frameworks are supported:
 | :------------: | ------------------------------------- |
 | **Ruby**       | RSpec, [Minitest][minitest], Cucumber |
 | **JavaScript** | Mocha, Jasmine                        |
-| **Python**     | Nose, PyTest                          |
+| **Python**     | Nose, PyTest, Django                  |
 | **Elixir**     | ExUnit                                |
 | **Go**         | Go                                    |
 | **Clojure**    | Fireplace.vim                         |
--- a/autoload/test/python/django.vim
+++ b/autoload/test/python/django.vim
@@ -1,0 +1,57 @@
+if !exists('g:test#python#django#file_pattern')
+  let g:test#python#django#file_pattern = '^test_.*\.py$'
+endif
+
+function! test#python#django#test_file(file) abort
+  if fnamemodify(a:file, ':t') =~# g:test#python#django#file_pattern
+    return !empty(findfile('manage.py', '.;'))
+  endif
+endfunction
+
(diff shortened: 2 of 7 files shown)


benjohnson:
Actually, sorry -- don't merge this yet. There's a problem with the TestSuite command.

README.md
```diff
@@ -184,6 +184,8 @@ the first available will be chosen, but you can force a specific one:
 let test#python#runner = 'pytest'
 " or
 let test#python#runner = 'nose'
+" or
+let test#python#runner = 'django'
```
janko-m: Is there the also a problem in detecting a Django project? Does every Django project have a `manage.py` file? Because I intended that here are listed only the ambigous runners.

benjohnson:
Ok, sorry about that. It seems I was not taking into consideration the case where the manage.py is in the current directory.

README.md
benjohnson: I've removed the reference. It's still slightly possible for it to be ambiguous if you, say, have the nose executable installed but are also testing a django app. But the django test runner is the default runner. So if you installed nose to test your django app with, this will currently do the "right" thing and find nose first.

janko-m: Hmm, I'm wondering if moving Django in `g:test#runners` to be the first is a better choice. If `manage.py` undeniable identifies a Django project, it's better that in that case the Django runner is being run, isn't it? In that case we wouldn't need the `test#python#runner` conditional, because these doesn't need to be checked for Django. What do you think?

benjohnson: Ok. I've thought about it and talked to some people and I think you're right. I was worried about a case where you're testing your Django app with nose, but a) that's not the default or majority at all, and b) our nose runner wouldn't be able to handle that case (tests are still run through manage.py).

So, I think the safe thing to do is exactly what you're suggesting. We detect if you're a Django project, and if you are you get the Django test runner. And if people want to get fancy with alternate runners for Django we can deal with that in a separate PR.

I've made the change.

janko-m:
https://github.com/janko-m/vim-test/commit/1b794fc040ceabd935258151f5ff2ca9b64e0e72

I cherry-picked your commit, just to change that the runner's name is `:DjangoTest` instead of `:Django`. This is kind of the convention that I have in vim-test, and the reason for this is to avoid overlapping with any potential vim-django plugins, which may want to define a `:Django` command to do something else (which makes more sense in a vim-django plugin than in vim-test).

Thank you so much! :heart:

Issue #56, opened 2015-09-06
Bug: Django runner incorrectly being selected, causing error


sloria:
I am using the latest `master` branch. When running any of the test commands in one my projects, I get the following error:

```
|| /var/folders/kd/9_jv37r94035mc9b94b3f_6c0000gn/T/vBLw9a9/37:56: permission denied: /Users/sloria/projects/flask-projects/manage.py
```

Django is not in my virtualenv and there is no manage.py file, so I'm not sure why the django runner is selected. Even if I set my test runner with:

```
let test#python#runner = 'pytest'
```

The error still happens.

janko-m:
@benjohnson Could you help me with this? It seems that `findfile()` is triggering the permissions error. The reason why you didn't just `glob('**/manage.py')` is because Django projects can have multiple applications, in which is common to `cd` in, so `manage.py` is higher in the directory structure, right? But can the application root be further than 1 folder down? As I see, in common Django directory structure manage.py is usually either in `manage.py` or `../manage.py`, no?

benjohnson:
Yeah. So in Django projects ./manage.py is always in the root directory of the project, but the idea was that your current working directory could be higher up the stack. So, if you're cd'd into myapp/someapp/tests and you run a test, we'd like it to pick up Django. What's worse, we need to know where the manage.py is so we can run tests from there.

Here's a few options off the top of my head:

1) We could just accept as a project constraint that you have to be cd'd into the root directory where manage.py is and avoid findfile() entirely, but I can see that affecting the way a lot of people use vim and Django.

2) We could check for the existence of the `django-admin` executable as a way to tell if we should run the test runner. That way we'd only be calling findfile() to get the path of ./manage.py if you actually have django installed (in the virtualenv).

I think giving (2) a shot is worth it. And if that still doesn't work we could go for (1). We also probably want to consider making sure that `test#python#runner` can skip the Django test entirely.

benjohnson:
[This branch](https://github.com/benjohnson/vim-test/tree/django-executable) has the change to checking for the executable instead of doing `findfile()` in case @sloria wants to give it a shot and see if it fixes his problem.

janko-m:
Test.vim is already built with the constraint of having to be cd'd in the root of the project. When I was building test.vim, I was thinking whether it should have this constraint, and I have found 2 reasons in favour:

* It's much easier to check for existence of files when you know Vim will be in the project root. Without this requirement each Vim runner would have to do complex directory traversal.
[...]

PR #57, opened 2015-09-06
Change django detection by assuming current directory is root (and has manage.py file).

benjohnson:
This commit does a few things:

1) The original Django runner was attempting to be too clever in trying to deal with cases where the current working directory wasn't root and therefore didn't have a  manage.py file we could run tests from. Instead, we went hunting for it up the file tree. vim-test as a basic constraint, however, requires the working directory to be root, so we can remove the cleverness and make the code much simpler and more performant.

2) Because we were searching up the paths looking for a manage.py file, we could encounter false positives in trying to find the file (see #56). This should partially resolve that problem by not searching all up the tree for a manage.py file. However, some other python frameworks could have a manage.py in root, so to cover those cases we're now also checking that django is installed by looking for the `django-admin` executable.

3) We were not allowing the django test runner to be overridden with the `test#python#runner` variable, which could be problematic for those with python installed who wish to use a different test runner or people just generally running into our check not being sufficient. So, we're now checking for that setting as well.

### DIFF
--- a/autoload/test/python/djangotest.vim
+++ b/autoload/test/python/djangotest.vim
@@ -4,7 +4,11 @@ endif
 
 function! test#python#djangotest#test_file(file) abort
   if fnamemodify(a:file, ':t') =~# g:test#python#djangotest#file_pattern
-    return !empty(findfile('manage.py', '.;'))
+    if exists('g:test#python#runner')
+      return g:test#python#runner == 'djangotest'
+    else
+      return (executable('django-admin') && executable(getcwd() . '/manage.py'))
+    endif
   endif
 endfunction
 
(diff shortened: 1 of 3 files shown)


janko-m:
I cherry-picked your commit: https://github.com/janko-m/vim-test/commit/3b3ef4fc4b777afefc218570f99747e294ad969b

I just changed that ./manage.py doesn't have to be executable in order for Django to be detected, that it's enough it exists. This would account for a solution where the user might have created a Django project manually (not through a generator), but forgot to make ./manage.py executable. It also isn't necessary to generate full path to ./manage.py with `getcwd()`, it's much cleaner to read if it's just `./manage.py test` :)

Thank you for the PR and your involvement!

PR #58, opened 2015-09-09
Added djangotest

akarl818:
I had to go through the source to find the name of the runner to use with Django, originally I used `'django'` but that didn't work, and the docs where not clear.


janko-m:
Good idea, I realized now that I should note identifiers for other test frameworks as well.

PR #63, opened 2015-09-30
javascript: alias "context" to "describe"

kevinburke:
Mocha's BDD test runner adds a `context` global, much like rspec's `context`:
https://mochajs.org/#bdd. Previously the namespace regex would miss this, which
would cause the grep to be incomplete. Updates the mocha regex to match the
word `context` in the same place as the word `describe`.

Adds tests that this new regex behaves as expected.

### DIFF
--- a/autoload/test/javascript.vim
+++ b/autoload/test/javascript.vim
@@ -1,4 +1,4 @@
 let test#javascript#patterns = {
   \ 'test':      ['\v^\s*it[( ]%("|'')(.*)%("|''),'],
-  \ 'namespace': ['\v^\s*describe[( ]%("|'')(.*)%("|''),'],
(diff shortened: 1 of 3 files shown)


kevinburke:
Apologies - I'm having trouble with the Ruby setup on my Mac. I couldn't get the test dependencies installed, so I'm not sure that the tests actually pass and cover the right code. I'd put maybe a 75% chance on them passing and covering the right thing.

kevinburke:
I'm also not sure of the history here - I believe that `context` used to match and a recent commit broke this behavior. I tried checking the commit history for `autoload/javascript` to see if this was deliberate, but it didn't appear to be.

janko-m:
Thanks for the update! And I really appreciate the tests. :heart:

PR #64, opened 2015-09-30
Added nearest support to PHPUnit using --filter argument

codeinabox:
Added nearest support to PHPUnit using --filter argument


janko-m:
Looks great, thanks! (although I find the GIF a bit disturbing)

codeinabox:
I may need to create another PR to change the order of --filter to come before the name of the PHP file. How would I do that?

janko-m:
Certainly. Simply change `return [a:position['file'], name]` to `return [name, a:position['file']]`.

PR #65, opened 2015-09-30
Fixed the order so that PHPUnit --filter option is before test file argument

codeinabox:
Fixed the order so that PHPUnit --filter option is before test file argument


codeinabox:
Thank you. Next time I'll test more to save me the embarrassment of creating a fix for my own PR.

janko-m:
It's not an embarassement at all, I'm happy that you're making the PRs :)

PR #66, opened 2015-09-30
[WIP] PHPSpec support

codeinabox:
Adds support for PHPSpec. All unit tests are passing but needs real world testing.


janko-m:
Awesome, thank you!

codeinabox:
I've tested this out on https://github.com/phpspec/phpspec own specs and it works all fine both for the suite and individual files.

PR #68, opened 2015-10-16
Add Maven runner for Java.

sookocheff:
This PR adds a Maven runner for Java JUnit tests.

### DIFF
--- a/autoload/test/java.vim
+++ b/autoload/test/java.vim
@@ -1,0 +1,4 @@
+let test#java#patterns = {
+  \ 'test':      ['\v^\s*public void (\w+)'],
+  \ 'namespace': ['\v^\s*public class (\w+)'],
+\}
--- a/autoload/test/java/maven.vim
+++ b/autoload/test/java/maven.vim
@@ -1,0 +1,39 @@
+if !exists('g:test#java#maven#file_pattern')
+  let g:test#java#maven#file_pattern = '\v^.*[Tt]est\.java$'
+endif
+
+function! test#java#maven#test_file(file) abort
+  return a:file =~? g:test#java#maven#file_pattern
+endfunction
+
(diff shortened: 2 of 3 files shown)


autoload/test/java.vim
```diff
@@ -0,0 +1,4 @@
+let test#java#patterns = {
+  \ 'test':      ['\v^\s*public void (\w+)'],
```
sookocheff: This could probably be made more specific by searching for the nearest `@Test` annotation and then finding the method name of the line below that. Suggestions on a regex to do that are welcome.

autoload/test/java/maven.vim
```diff
@@ -0,0 +1,39 @@
...
+function! test#java#maven#build_position(type, position) abort
+  if a:type == 'nearest'
+    let name = s:nearest_test(a:position)
+    if !empty(name)
+      return [name]
+    else
+      return [filename]
+    endif
+  elseif a:type == 'file'
+    let strip_extension = split(a:position['file'], "\.java")[1]
```
sookocheff: Here, we have a file like `src/test/java/main/MyTest.java`. I need to reduce this to `MyTest`. I'm not sure if there is a better way to do this with vimscript.

autoload/test/java.vim
janko-m: It's probably fine, we'll see how it goes.

janko-m:
Awesome, thank you!

autoload/test/java/maven.vim
janko-m: Hehe, sorry, merged it too soon. `fnamemodify(a:position['file'], ':t:r')` should do what you want.

sookocheff: Fixed here: https://github.com/janko-m/vim-test/pull/69/files

PR #69, opened 2015-10-16
Fix bug with filename computation.

sookocheff:
Fix a bug introduced during some cleanup.

### DIFF
--- a/autoload/test/java/maven.vim
+++ b/autoload/test/java/maven.vim
@@ -7,6 +7,8 @@ function! test#java#maven#test_file(file) abort
 endfunction
 
 function! test#java#maven#build_position(type, position) abort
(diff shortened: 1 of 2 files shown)


autoload/test/java/maven.vim
```diff
@@ -15,8 +18,6 @@ function! test#java#maven#build_position(type, position) abort
       return [filename]
```
sookocheff: Bug here: filename may not be defined if we can't find a nearest match.

janko-m:
Looks good. Could you just squash it into a single commit?

sookocheff:
@janko-m Squashed. Thank you.

janko-m:
Great, thanks again :)

PR #71, opened 2015-10-22
Fixed PHPUnit nearest to handle tests with arguments

codeinabox:


janko-m:
Awesome, thank you! Also, great GIF.

PR #72, opened 2015-10-28
Handling when `TestFile` is called on a Go test in a subdirectory

codeinabox:

### DIFF
--- a/autoload/test/go/gotest.vim
+++ b/autoload/test/go/gotest.vim
@@ -12,7 +12,8 @@ function! test#go#gotest#build_position(type, position) abort
     if !empty(name) | let name = '-run '.shellescape(name, 1) | endif
     return [name]
   elseif a:type == 'file'
-    return []
+    let path = fnamemodify(a:position['file'], ':h')
+    return path != '.' ? ['./' . path . '/...'] : []
   else
     return ['./...']
   endif
--- a/spec/fixtures/gotest/mypackage/normal_test.go
+++ b/spec/fixtures/gotest/mypackage/normal_test.go
@@ -1,0 +1,7 @@
+package mypackage
+
+import "testing"
+
+func TestNumbers(*testing.T) {
+	// assertions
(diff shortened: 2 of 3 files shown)


codeinabox:
@janko-m I've written a scenario to cover the scenario that is causing my tests in a project not to run but what is the best way to implement this?

janko-m:
I'm not that familiar with Go, so I'm not sure. Did I get it right that this test is currently passing, but a test file on your project is still not being run?

codeinabox:
Currently it doesn't support running tests in a subdirectory when doing `TestFile` as `go test` only runs tests in the current directory. I've added a new test to indicate how it should behave but not sure how to implement it as my Vimscript is not so good.

janko-m:
Aha, now I get it. You probably want `fnamemodify()`, it has a lot of options, so you should probable be able find what you need there.

codeinabox:
Done. This passes the tests though maybe can be written more elegantly

autoload/test/go/gotest.vim
```diff
@@ -12,7 +12,8 @@ function! test#go#gotest#build_position(type, position) abort
     if !empty(name) | let name = '-run '.shellescape(name, 1) | endif
     return [name]
   elseif a:type == 'file'
-    return []
+    let path = fnamemodify(a:position['file'], ':h.')
+    return path != '.' ? ['./' . path . '/...'] : []
```
janko-m: Hmm, what does the dot mean? Is it maybe better to write it like this?

```vim
let path = fnamemodify(a:position['file'], ':h')
return !empty(path) ? ['./' . path . '/...'] : []
```

codeinabox: I don't need the dot in `fnamemodify` though I do need to check if the path is equal to "." which is the current directory. Have updated my PR accordingly.

janko-m:
Great, thanks for the patch!

PR #75, opened 2015-11-10
Can run nearest phpunit test marked with @test annotation

mjacobus:
The simplest solution was implemented, due to my lack of VimL knowledge. Now it works, but not because of the `@test` annotation, but because it is the nearest method.

The way it is it will consider any method as a test method. That may be considered a bug.

Of course it can be improved, but IMO it is better to see PHPUnit's message "no tests executed!" when cursor is near a non-test method than not being able to run a single method when it is an annotated method.

### DIFF
--- a/autoload/test/php/phpunit.vim
+++ b/autoload/test/php/phpunit.vim
@@ -39,7 +39,12 @@ function! test#php#phpunit#executable() abort
 endfunction
 
 function! s:nearest_test(position)
-  let patterns = {'test': ['\vpublic function (test\w*)\('], 'namespace': []}
+  let patterns = {
+    \ 'test': [
+    \    '\vpublic function (test\w*)\(',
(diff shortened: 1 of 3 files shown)


autoload/test/php/phpunit.vim
```diff
@@ -39,7 +39,12 @@ function! test#php#phpunit#executable() abort
 endfunction
 
 function! s:nearest_test(position)
-  let patterns = {'test': ['\vpublic function (test\w*)\('], 'namespace': []}
+  let patterns = {
+    \ 'test': [
+    \    '\vpublic function (test\w*)\(',
```
mjacobus: This line is no longer needed, since the following line can also cover the test-prefixed methods.

janko-m:
Sorry for the late response, thanks for the patch, I agree it makes sense.

PR #77, opened 2015-11-19
Added djangonose plugin

peterdemin:
### Preface
I test django project with django-nose plugin.
It allows to enjoy all benefits of nosetests, but have different test location pattern, then unittest.
### Patch
I copied djangotest.vim, and changed line 20: replaced separator with ':' (was '.')
This is the only change I made to djangotest.vim

Looks like I am not following DRY principle here ;) but I am very new to vim plugins programming and don't know how to do it correctly.

### DIFF
--- a/autoload/test/python/djangonose.vim
+++ b/autoload/test/python/djangonose.vim
@@ -1,0 +1,50 @@
+if !exists('g:test#python#djangotest#file_pattern')
+  let g:test#python#djangotest#file_pattern = '^test_.*\.py$'
+endif
+
+function! test#python#djangotest#test_file(file) abort
+  if fnamemodify(a:file, ':t') =~# g:test#python#djangotest#file_pattern
+    if exists('g:test#python#runner')
+      return g:test#python#runner == 'djangotest'
+    else
+      return filereadable('python manage.py') && executable('django-admin') 
+    endif
+  endif
+endfunction
+
+function! test#python#djangotest#build_position(type, position) abort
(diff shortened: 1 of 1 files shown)


janko-m:
Thanks for the addition. One thing is missing though feature-wise, it should detect whether the Django project is using the "django-nose" plugin, and in `test_file()` function only return true in that case. This is becase vim-test iterates over all runners and runs this function, and then sets the runner to the first one that returns true (so in this case either djangonose or djangotest will always be used, depending on which one is first in the list).

In VimScript it's a bit more difficult to be DRY since there are less constructs for code sharing, but this is what I would call "coincidental duplication", in the way that it looks the same, but it may later get different behaviour because they are still different runners.

peterdemin:
AFAIK there is no fast way to detect django-nose usage - it can configured in django settings module, in our project loading settings takes about 4 seconds. I think it is acceptable to use djangotest by default, but allow to configure djangonose runner in .vimrc

janko-m:
Hmm, does this settings module always live in the same place? Couldn't we just do `system('cat path/to/settings_module.py') =~ 'django-nose'`?

peterdemin:
Not really. In fact we have 5 settings files and some of them include others. Active settings file path is configured through command-line parameters to manage.py or environment variable.
Even if 'django-nose' is inside chosen settings file, it can be flagged out by conditions.

janko-m:
After some thinking, I've decided that it's better to just keep it in the same runner. Thanks a lot for making the PR with the solution.

PR #81, opened 2015-11-30
Don't clear the screen before running tests

nixon:
Provide a variable that can be used to prevent the screen from being
cleared before each test run.  Set `g:test#preserve_screen` to non-zero
to disable the clear screen functionality.

By default, the existing clear screen behavior is kept.

### DIFF
--- a/autoload/test/strategy.vim
+++ b/autoload/test/strategy.vim
@@ -69,7 +69,11 @@ function! s:pretty_command(cmd) abort
   let clear = !s:Windows() ? 'clear' : 'cls'
   let echo  = !s:Windows() ? 'echo -e '.shellescape(a:cmd) : 'Echo '.shellescape(a:cmd)
 
-  return join([l:clear, l:echo, a:cmd], '; ')
+  if !exists('g:test#preserve_screen') || !g:test#preserve_screen
+    return join([l:clear, l:echo, a:cmd], '; ')
+  else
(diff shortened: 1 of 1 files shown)


janko-m:
This option definitely seems useful, but why do you want this behaviour? I know it's probably a matter of preference.

nixon:
I often use Ctrl-Z to suspend vim and it became annoying to have what I was working on disappear after running tests.  :)

janko-m:
Gotcha, thanks a lot for the PR!

laserlemon:
@janko-m, @nixon: Am I right in assuming that when the screen is _preserved_, that the `echo` command in `pretty_command` is no longer needed/wanted? It looks like the purpose is to display the command being run, assuming the screen has just been cleared.

If that sounds right, I'd like to submit a PR to remove the `echo` command when the screen is preserved. Thank you!

janko-m:
@laserlemon I think it's still useful to have vim-test display the command that's being run, even when the screen is preserved. But if you prefer not to have it, I'd be willing to accept a PR that adds a configuration option to disable it, e.g. `test#echo_command` or `test#no_echo`.

laserlemon:
Sounds good, thank you! For reference, here's a shot of my tmux with echo and preserving the screen:

screen shot 2018-05-30 at 2 29 25 pm

I'll open a PR as I'm able. Thank you!

PR #82, opened 2015-12-25
Add support for PHPUnit custom executable

vladbalmos:
Hello,

My use case is running PHPUnit from inside a docker container. To accomplish this I create a bash script which wraps the docker command and parameters and call that instead of the phpunit executable, something like:

    ./run-tests.sh tests/

Unfortunately I do not have the necessary skills to write a test for the modification but I do hope you will consider merging it.

### DIFF
--- a/autoload/test/php/phpunit.vim
+++ b/autoload/test/php/phpunit.vim
@@ -29,6 +29,9 @@ function! test#php#phpunit#build_args(args) abort
 endfunction
 
 function! test#php#phpunit#executable() abort
+  if exists('g:test#php#phpunit#executable')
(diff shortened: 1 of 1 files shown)


janko-m:
It so happens that test.vim already has this feature, and works automatically for all test runners, you just assign the variable with exactly that name ;)

vladbalmos:
Ah... that makes sense! :)

janko-m:
Ok, I'll assume this worked, so I'll close this.

```vim
let g:test#php#phpunit#executable = ''./run-tests.sh"
```

PR #85, opened 2016-01-23
Add support for perl's prove runner

lucastheisen:
Add support for perl's core [prove](http://perldoc.perl.org/TAP/Harness.html#prove) runner

### DIFF
--- a/autoload/test/perl/prove.vim
+++ b/autoload/test/perl/prove.vim
@@ -1,0 +1,27 @@
+if !exists('g:test#perl#prove#file_pattern')
+      let g:test#perl#prove#file_pattern = '\vt?/.*\.(t|pl)$'
+endif
+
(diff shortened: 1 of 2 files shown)


janko-m:
Thanks for the pull request!

I just have one question about the file pattern, is there a reason that you made the "t" directory optional, and that you allow the .pl extension for tests? Are sometimes tests written outside of the "t" directory and with a .pl extension instead of a .t extension? It just seems that your pattern currently matches any Perl file.

Because one of test.vim's features is based on the recognition between application and tests files, so that if you run `:TestFile` in an application file, it recognizes that it's not a test file and instead runs the last test. It's a small feature, and you can instead just run `:TestLast` explicitly, but that does require a small mental overhead in my experience.

lucastheisen:
@janko-m, how about now?

PR #87, opened 2016-01-28
Mention binstubs for running Ruby tests

stevenharman:


janko-m:
Thanks, I agree, it's a good idea to mention this.

PR #88, opened 2016-02-01
Update README

derekprior:
1. We can be reasonably certain that people ran tests from their editor
before Gary Bernhardt.

2. Criticisms of other projects as having "bad designs" or "unclear
ideas" are unsubstantiated and, frankly, unkind. They don't bolster the
case for using `test.vim`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,10 +23,11 @@ Currently the following testing frameworks are supported:
 
 ## Idea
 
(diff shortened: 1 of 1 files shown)


janko-m:
Thank you for the update. I completely agree with both of your changes and arguments.

derekprior:
:beers:

PR #89, opened 2016-02-12
add support for prove test arguments

lucastheisen:
Added support for supplying arguments to tests themselves (not the test runner).  Also added a couple base options

### DIFF
--- a/autoload/test/perl/prove.vim
+++ b/autoload/test/perl/prove.vim
@@ -18,7 +18,17 @@ endfunction
 
 " Returns processed args (if you need to do any processing)
 function! test#perl#prove#build_args(args)
-    return a:args
+    let args = split(a:args[0], "::")
+
+    if test#base#no_colors()
+        let args[0] = ['--nocolor'] + args[0]
+    endif
+
+    if test#base#verbose()
+        let args[0] = ['--verbose']  + args[0]
+    endif
+
+    return [args[0], a:args[1], '::', args[1]]
 endfunction
 
 " Returns the executable of your test runner
--- a/autoload/test/perl/prove.vim
+++ b/autoload/test/perl/prove.vim
@@ -18,17 +18,25 @@ endfunction
 
 " Returns processed args (if you need to do any processing)
 function! test#perl#prove#build_args(args)
-    let args = split(a:args[0], "::")
-
-    if test#base#no_colors()
-        let args[0] = ['--nocolor'] + args[0]
+    if len(a:args) == 1 
+        return a:args
+    else
+        let args = split(a:args[0], '::')
+        
+        if test#base#no_colors()
+            let args[0] = ['--nocolor'] + args[0]
+        endif
+        
+        if test#base#verbose()
+            let args[0] = ['--verbose']  + args[0]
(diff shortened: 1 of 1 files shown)


lucastheisen:
hold on, this has a bug...  let me fix

lucastheisen:
Merged in the changes from upstream...  Should be good to go now.

janko-m:
Thank you for the PR. Since I'm not familiar with Prove, it's a bit difficult for me to figure out what is the intended behaviour of these changes. Also, `test#base#verbose()` doesn't exist. Do you think you could add some tests around this?

lucastheisen:
The main purpose of this modification is to add the ability to supply arguments to the tests themselves.  Prove sends anything after the arisdottle (`::`) to the tests that it runs.  I use this to supply the tests with verbosity level when also supplying `--verbose` which is an argument to prove itself.  The main change was due to the fact that `test#perl#prove#build_args` always supplies the test file name last and the arisdottle _MUST_ be after the file list to prove.  

As to there not being a `test#base#verbose()`, i _incorrectly_ assumed that was just a way of allowing users to specify values in their `.vimrc` for that parameter.  That said, it can be removed.  My standard use case is open the test I am working on and `:TestFile --verbose :: --log trace`, so I wasn't even using it anyway.

As to adding tests, sad to say, I don't know how...  I could probably figure it out as soon as i get a second.

janko-m:
Ok, no worries, I'll add the tests. Thanks for the explanation :)

janko-m:
Ok, I cherry-picked your change to master, and fixed some bugs in the implementation and added tests. Thank you!

lucastheisen:
The version in `master` does not have the split on `::` which was the purpose of this issue/pull request.   I cannot use this plugin as it is currently implemented.  Is there a reason you removed the 
```
  if len(a:args) == 2
    let args = split(a:args[0], '::')

    if test#base#no_colors()
      let args[0] = ['--nocolor'] + args[0]
    endif

    if test#base#verbose()
      let args[0] = ['--verbose']  + args[0]
    endif

    let out_args = [args[0], a:args[1]]
    if len(args[1])
      let out_args = out_args + ['::', args[1]]
    endif
  endif
```
part?

janko-m:
I'm surprised the test arguments don't work properly, I must have missed something in the implementation. The reason I rewrote your code is because I don't think any code in `test#perl#prove#build_args()` should depend on the arguments length. For example, if there are any options set up in `test#perl#prove#options`, the arguments length would automatically be 3, and your code only checked if they're of length 2. Also, I've realized that arguments passed to `:Test*` commands should be recorded as an array of arguments (and not string), in order to be consistent with other arguments, so `test#perl#prove#build_args()` can count that all arguments are split as an array, and is able to manipulate it the same regardless of the source of these arguments. In that case the number of arguments would be the number of CLI arguments.

What exactly doesn't work? I already wrote some [tests](https://github.com/janko-m/vim-test/blob/ec07294b8dcdd16e13e1b77a84d483f58cbb9738/spec/prove_spec.vim#L38-L46) for test arguments with `:TestFile`.

PR #91, opened 2016-02-19
Feature/djangotest fixes

DXist:
* fix default test discovery pattern
* fix manage.py file check


janko-m:
Wow, I don't know how I missed those. Thanks for the patch!

PR #96, opened 2016-03-15
Use --no-ansi with PHPSpec if strategy doesn't support colors

codeinabox:


janko-m:

PR #99, opened 2016-04-05
Test runner executable search depth and integration with phpunit executable search

mmikitka:
Search up the directory tree from a given test file for the phpunit executable until a match is found or g:test#executable_search_depth is reached.

Including @codeinabox since he has committed a few PHP-related patches.

### DIFF
--- a/autoload/test/php/phpunit.vim
+++ b/autoload/test/php/phpunit.vim
@@ -38,6 +38,38 @@ function! test#php#phpunit#executable() abort
   endif
 endfunction
 
+function! s:search_local_executable(dir) abort
+  " Look for phpunit in Composer's vendor/bin and Symfony's bin directories
+  echo a:dir
+  if filereadable(a:dir . '/vendor/bin/phpunit')
+    return a:dir . '/vendor/bin/phpunit'
+  elseif filereadable(a:dir . '/bin/phpunit')
+    return a:dir . '/bin/phpunit'
+  else
+    return ''
+  endif
+endfunction
+
+function! test#php#phpunit#executable() abort
+  let phpunit_path = ''
+  let search_count = 0
+  let search_base_dir = expand("%:p:h")
+  while search_count < g:test#executable_search_depth
+    let phpunit_path = s:search_local_executable(search_base_dir)
+    if phpunit_path != ''
+        break
+    endif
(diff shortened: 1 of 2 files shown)


codeinabox:
Great idea. Once it has found it does it keep the path in memory or need to look it up each time? @janko-m or would it be better to make the path confgurable as an option?

mmikitka:
It doesn't store the path in memory, but that can be done.

I don't think that we want to make the path a configuration option, since
it is dynamically located relative to the position of the test file and the
phpunit binary.

matt

janko-m:
Hmm, can the location of the `phpunit` executable really be that dynamic? Isn't it easier to just set the executable per project if the executable is in a non-standard place:

```rb
let test#php#phpunit#executable = 'path/to/phpunit'
```

Actually, is it rather tha Composer's or Symphony's `phpunit` executable is actually located in the project root, and this PR was an attempt to cover the case where your current working directory (directory you're cd'd into) isn't the project root? Because test.vim's design mandates being cd'd into the project root.

codeinabox:
I agree with @janko-m, the location is based I've experienced working on both Symfony and projects using Composer. Since adding support last year I haven't come across a project where it's in a different location, though I have installed phpunit globally with homebrew.

mmikitka:
@janko-m Yes to the latter reason for making this change i.e., the working
directory is outside of the project root. In my case, Vim is configured to
change the cwd to the file in the active buffer, so the location of phpunit
is relative as such.

janko-m:
Yes, unfortunately, you cannot use test.vim with the `autochdir` setting, it expects you to always be cd'd into the project root. Handling `autochdir` would bring too much complexity to the plugin, each test runner would have to handle this separately. Here are some of the main issues that would arise:

* checking file patterns wouldn't work, because some patterns rely on the directory name
* every call to `filereadable()`, `executable()` and `isdirectory()` would have to be wrapped to traverse the filesystem
* executing the last file test would work only if you're executing the command from a buffer of a file which is in the same directory as the test file

The 1st and 3rd point could be mitigated by switching to absolute paths, though, but the 2nd point is really tricky. So, the only way to make test.vim work properly is to disable the `autochdir` Vim setting, or whatever is automatically changing the working directory.

Issue #101, opened 2016-04-07
Perl prove fails TestSuite if using a non-standard folder for test files


lucastheisen:
I have a strange situation in which i am writing perl code in a maven style project folder structure.  This does not follow the default `t/*.t` pattern.  I can easily update the file_pattern:
```
let g:test#perl#prove#file_pattern = '\v^src/test/perlt/.*\.t$'
```
Which allows `:TestFile` to work, but when running `:TestSuite` it calls `prove` without arguments.  When `prove` is called without arguments, it assumes `t\*.t`.

lucastheisen:
My current workaround is to use a project specific `.vimrc`:
```
let g:test#perl#prove#test_dir = 'src/test/perlt'
let g:test#perl#prove#file_pattern = '\v^' . g:test#perl#prove#test_dir . '.*\.t$'

function! test#perl#prove#build_args(args)
    let args = []
    let test_args = []

    if empty(a:args)
        let args = [g:test#perl#prove#test_dir]
    else
        echom 'NON empty a:args'
        for idx in range(0, len(a:args) - 1)
            if a:args[idx] == '::' || !empty(test_args) && !test#base#file_exists(a:args[idx])
                call add(test_args, a:args[idx])
            else
                call add(args, a:args[idx])
            endif
        endfor
    endif

    let args = args + test_args

    if !empty(filter(copy(args), 'isdirectory(v:val)'))
        let args = ['--recurse'] + args
    endif

    if test#base#no_colors()
        let args = ['--nocolor'] + args
    endif

    return args
endfunction
```
I would submit a pull request if I knew of a more standard approach for setting default test folder, or `:TestSuite` command or something...  Do you have a standard way to approach this issue?

janko-m:
Ok, I think it's better to make the default file pattern just `\.t$`, so that tests can be put in any folder. As for making `:TestSuite` work, vim-test allows setting options for each granularity, so you can do:

```vim
let test#perl#prove#options = {'suite': 'src/test/perlt'}
```

Then `test#perl#prove#build_args` should automatically add the `--recurse` option. Let me know if this works.

janko-m:
Actually, I think I shouldn't change the default file pattern to `\.t$`, as it could introduce potential backwards incompatibility with another test runner which might be using the same extension, and this is probably a non-standard setup anyway, so it makes more sense to override the file pattern by hand like you did.

lucastheisen:
[...]

PR #102, opened 2016-04-07
Improved prove support

lucastheisen:
Added support for non-standard test directory (resolving [#101](Perl prove fails TestSuite if using a non-standard folder for test files )).  Added back support for test arguments that was removed by cherry pick (resolving [this regression](https://github.com/janko-m/vim-test/pull/89#issuecomment-207086204)).


lucastheisen:
Ignore this pull request... looks like there were other changes that made the assumption of 2 arguments incorrect...  I will refactor and create a new pull request.

Issue #105, opened 2016-04-09
JavaScript's Intern framework support


aaalsaleh:
Any chances of supporting JavaScript's [Intern](https://theintern.github.io) framework?

janko-m:
Merged in https://github.com/janko-m/vim-test/pull/108.

PR #107, opened 2016-04-22
Add support for JavaScript's Intern framework

aaalsaleh:
This pull request adds Intern support (#105) with the following features:
- Supports BDD/TDD/Object interfaces.
- Supports both unit and functional tests.
- Supports `TestNearest`, `TestFile`, and `TestSuite`.
- Allows specifying Intern's test runner (`intern-client` vs `intern-runner`) through `g:test#javascript#intern#test_runner`.
- Allows specifying Intern's configuration module (`tests/intern`) through `g:test#javascript#intern#config_module`.

Things to note:
- It is impossible to implicitly detect which test runner to use. Therefore, the default is `intern-client`. However, the user can change it through `g:test#javascript#intern#test_runner` to `intern-runner`.
- It is impossible to implicitly detect whether the test is a unit test or a functional test without assumptions. Intern recommends using the file structure `tests/unit/*` and `tests/functional/*` to separate the two. Therefore, this recommendation was used to differentiate between the two implicitly. Anything under `test/functional/*` is considered functional test, and anything else is considered regular unit test.
- Even though Intern recommends sticking with the file `tests/intern` as the default config module, the users can still choose any other file as the config module. Therefore, the user must set  `g:test#javascript#intern#config_module` appropriately if they choose to use anything other than the default `tests/intern`.
- ~~Even though I wrote the test cases in `spec/`, I couldn't manage to run it as I couldn't setup `vim-flavour` successfully. I did however test everything manually.~~ It actually passed the tests :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,6 +82,12 @@ let test#strategy = "dispatch"
 | **Terminal.app**                | `terminal` | Sends test commands to Terminal (useful in MacVim GUI).                          |
 | **iTerm2.app**                  | `iterm`    | Sends test commands to iTerm2 >= 2.9 (useful in MacVim GUI).                     |
 
+In addition to setting a strategy globally, you can also set one per command:
+
+```
+:TestFile -strategy=neovim
+```
+
 Some strategies clear the screen before executing the test command, but you can
 disable that by setting `g:test#preserve_screen`:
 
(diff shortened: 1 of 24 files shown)


aaalsaleh:
The failure in mocha test was due to me changing `[( ]` to `[(]`. I didn't think of CoffeeScript. It is now fixed.

PR #108, opened 2016-04-23
Add support for JavaScript's Intern framework (#2)

aaalsaleh:
This pull request adds Intern support (#105) with the following features:
- Supports BDD/TDD/Object interfaces.
- Supports both unit and functional tests.
- Supports `TestNearest`, `TestFile`, and `TestSuite`.
- Allows specifying Intern's test runner (`intern-client` vs `intern-runner`) through `g:test#javascript#intern#test_runner`.
- Allows specifying Intern's configuration module (`tests/intern`) through `g:test#javascript#intern#config_module`.

Things to note:
- It is impossible to implicitly detect which test runner to use. Therefore, the default is `intern-client`. However, the user can change it through `g:test#javascript#intern#test_runner` to `intern-runner`.
- It is impossible to implicitly detect whether the test is a unit test or a functional test without assumptions. Intern recommends using the file structure `tests/unit/*` and `tests/functional/*` to separate the two. Therefore, this recommendation was used to differentiate between the two implicitly. Anything under `test/functional/*` is considered functional test, and anything else is considered regular unit test.
- Even though Intern recommends sticking with the file `tests/intern` as the default config module, the users can still choose any other file as the config module. Therefore, the user must set  `g:test#javascript#intern#config_module` appropriately if they choose to use anything other than the default `tests/intern`.

### DIFF
--- a/autoload/test/javascript.vim
+++ b/autoload/test/javascript.vim
@@ -1,4 +1,11 @@
 let test#javascript#patterns = {
-  \ 'test':      ['\v^\s*it[( ]%("|'')(.*)%("|''),'],
-  \ 'namespace': ['\v^\s*%(describe|context)[( ]%("|'')(.*)%("|''),'],
+  \ 'test': [
+    \ '\v^\s*%(%(bdd\.)?it|%(tdd\.)?test)\s*[( ]\s*%("|'')(.*)%("|'')\s*,',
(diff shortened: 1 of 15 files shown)


janko-m:
This is amazing. It's so great to receive a PR for a testing framework from someone who obviously knows it very well. I can see that you've put a lot of thought into the implementation, and I really appreciate the extensive tests. Thank you! ❤️

Issue #84, opened 2016-01-19
Add Always Use Absolute Path Option?


bradurani:
I've run into an issue with relative paths. We have a Rails project with internal engines and internal gems. To run the specs in the engines and gems you have to cd into the engines directory in the terminal.

That causes a problem the first time you run a spec because the spec doesn't run from the right directory, but it's not a big deal because you can cd in the terminal and rerun and be good for the rest of the day.

There's a bigger problem however in that after you've done that, running the specs from vim-test uses a path relative to the project root. Do you think it's worthwhile adding a config variable to vim-test that will force it to send an absolute path to the terminal for every spec run?

janko-m:
Yes, adding that option makes a lot of sense, I'll add it later today.

janko-m:
I reread your issue description one more time, and I realized that I don't fully understand the problem. Vim-test doesn't create paths relative to the project root, it creates them relative to the working directory. So when you `cd` in and out of the engine, vim-test should always generate correct paths, since it should always be relative to the directory where you `cd`ed.

bradurani:
Ahhh... ok

For my workflow to function, I have to have Vim in the project root and the shell in the engine root. So I have to run the spec once, which errors, go to the shell, cd into the engine, go back to vim and run the spec again. That's when vim-test sends a relative path to the shell that doesn't match up.

janko-m:
Aha, I think I understand now, you're probably using a strategy where you send commands to the other shell window, the one which is `cd`ed into the engine root. Ok, I will add this option the first thing tomorrow.

bradurani:
Thank you so much!!!!

PR #109, opened 2016-04-23
CD into CWD before executing a runner in external terminals

aaalsaleh:
Currently, the external terminal need to be CDed into Vim's CWD manually for the test runner to execute successfully. This tiny PR automatically CD into Vim's CWD before executing any test runner in external terminals. Test with Terminal.app successfully, and should work with other terminals.

### DIFF
--- a/autoload/test/strategy.vim
+++ b/autoload/test/strategy.vim
@@ -54,12 +54,13 @@ endfunction
 
 function! s:pretty_command(cmd) abort
   let clear = !s:Windows() ? 'clear' : 'cls'
+  let cd = 'cd ' . shellescape(getcwd())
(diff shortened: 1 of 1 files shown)


janko-m:
I agree with this change, it's very useful to be able to `cd` to any directory in the Terminal, and just have test.vim bring you back to the project root.

Just as a note to others, I'm aware from https://github.com/janko-m/vim-test/issues/84 that some people might be relying on the shell being `cd`-ed to a different directory than Vim's working directory. However, after giving it some thought, I don't think test.vim should allow this, because file regexes might not work properly (e.g. file path is `anotherapp/test/foo.js`, but the regex checks for `^test/.*\.js`). Also, if that subdirectory is a subproject with it's own structure (like it is in case of https://github.com/janko-m/vim-test/issues/84 a Rails engines), file detection methods like `filereadable()` will not work properly because they'll look in the wrong directory (whole project's root instead of subproject's root).

Issue #110, opened 2016-04-24
Possible issue for some runners when TestNearest is using regex polyfill


aaalsaleh:
Some runners only allow running a single test (or subgroup of tests) by specifying a regex pattern matching its description instead of a specific line or class/method. For example, in Intern and Mocha, the only way to run a single test is to pass a grep with a regex of one or more suite (describe/context) and a test (it). There is no issue with vim-test finding the correct name of the suite(s) and test. However, there is an issue with the generated regex for grep'ing these tests when attempting a `TestNearest`. Currently, the generated regex in some of these framework is never prefixed with `^` or postfixed with `$` depending on the case. Take this test sample as an example:
```javascript
it('play and stop', function() {
  ...
});

it('play', function() {
 ... // cursor here
});

describe('playing', function() {
 ...
});
```
Suppose the cursor is within the test "play". If we run `TestNearest`, and the current runner generates regex to specify the test, then it would be "play". Since it is a regex pattern, it would not only match and run the test "**play**", but also the test "**play** and stop" and even the whole suite "**play**ing" and any of its subtests. Obviously, this is not the wished behaviour of `TestNearest` regex polyfill.

To solve this, `^` and `$` must be used carefully. If the cursor is near a test (possibly inside suite/describe), then it must be prefixed with `^` and postfixed with `$`. This would run only the exact pattern (a single match). In this case there would be both namespace(s) and a test name. However, if the cursor is near a suite/describe (not a test), then it must be prefixed with `^` only (without `$`). This would allow running the nearest suite/describe along with its subtests. In this case there would be only namespace(s) without a test name.

I would probably issue a PR to fix this issue for Intern (also add the QUnit interface that I missed to support on the way). Similar thing could be made for Mocha and other "regex polyfill"-based runners.

janko-m:
Thank you again for making the PRs, they have now both been merged. But I would like to leave this issue open to remind me to solve this for all other test runners that work with regexes (e.g. Jasmine should probably get it, and Ruby's Minitest has some inconsistencies).

aaalsaleh:
[...]

PR #111, opened 2016-04-26
Add support for Intern's QUnit interface and improve nearest test regex

aaalsaleh:
**Improvements:**
- Add support for Intern's QUnit interface.
- Move Intern's namespace and test pattern away from `javascript.vim` to `intern.vim`. Some of the patterns are specific to Intern, and should not be globalized for other javascript test frameworks. Since Intern can support other testing interfaces, the default javascript patterns were added as well.
- Improve `TestNearest` regex polyfill (#110). If the nearest test detected is:
  - Namespace(s) only, then `^` will be prefixed only. This allows running anything starting with the namespace(s) pattern only (all sub-namespaces and sub-tests).
  - Test inside namespace(s), then `^` will be prefixed, and `$` will be postfixed. This allows running anything exactly matching the test pattern inside the namespace(s).
  - Test without any suite, then `$` will be postfixed only (QUnit only, see limitation). This allows running test(s) ending with the test pattern.

**QUnit Limitation:**
QUnit grouping of tests under namespaces/module is weird. It takes a linear approach similar to this:
```javascript
QUnit.module('Suite #1');
QUnit.test('Part of Suite #1', ...);
...
QUnit.module('Suite #2');
QUnit.test('Part of Suite #2', ...);
...
```
Since vim-test scans the scopes incrementally for test then namespace(s), if the cursor is near a test, then that test will be detected, but the namespace/module will not be detected. This is due to the fact that both of them are on the same scope. This is why I decided to postfix with `$` (without `^`), if only a test was detected (without namespaces). QUnit is the only interface suffering from such limitation, and all other interfaces will function properly.

QUnit has an alternative way of using scoped namespace(s)/module(s), but it is not supported by Intern's implementation currently. If it get supported in the future, then there is no need to do any changes on the patterns, as it should detect it correctly since it is scoped.

### DIFF
--- a/autoload/test/javascript.vim
+++ b/autoload/test/javascript.vim
@@ -1,11 +1,4 @@
 let test#javascript#patterns = {
-  \ 'test': [
-    \ '\v^\s*%(%(bdd\.)?it|%(tdd\.)?test)\s*[( ]\s*%("|'')(.*)%("|'')\s*,',
-    \ '\v^\s*%("|'')(.*)%("|'')\s*:\s*function\s*[(]'
-  \],
-  \ 'namespace': [
-    \'\v^\s*%(%(bdd\.)?describe|%(tdd\.)?suite|context)\s*[( ]\s*%("|'')(.*)%("|'')\s*,',
-    \ '\v^\s*%("|'')(.*)%("|'')\s*:\s*[{]',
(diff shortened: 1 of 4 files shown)


janko-m:
I completely agree with these changes. Amazing work with these regexes, and making JS runners more robust, thank you! ❤️

PR #112, opened 2016-04-26
Improve Mocha's TestNearest regex polyfill

aaalsaleh:
Improve Mocha's TestNearest regex polyfill (#110). If the nearest test detected is:
- Namespace(s) only, then `^` will be prefixed only. This allows running anything starting with the namespace(s) pattern only (all sub-namespaces and sub-tests).
- Test inside namespace(s), then `^` will be prefixed, and `$` will be postfixed. This allows running anything exactly matching the test pattern inside the namespace(s).
- Test not inside any namespace, then `$` will be postfixed only. This allows running test(s) ending with the test pattern.
- Nothing, then an empty string will be returned.

I've updated `mocha_spec.vim` to reflect the new changes. I've also tested the new changes with both vim-flavor and mocha installation quickly.

### DIFF
--- a/autoload/test/javascript/mocha.vim
+++ b/autoload/test/javascript/mocha.vim
@@ -10,7 +10,7 @@ function! test#javascript#mocha#build_position(type, position) abort
   if a:type == 'nearest'
     let name = s:nearest_test(a:position)
     if !empty(name)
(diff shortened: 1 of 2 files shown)


janko-m:
This is great, thank you!

PR #113, opened 2016-04-27
Run Go nearest tests from their directories

gylaz:
In Go, additional packages can live in subdirectories. In order to run
them properly, the subdirectory needs to be specified.

This fix ensures that the nearest test (by name) is run in the
appropriate directory.

### DIFF
--- a/autoload/test/go/gotest.vim
+++ b/autoload/test/go/gotest.vim
@@ -7,15 +7,16 @@ function! test#go#gotest#test_file(file) abort
 endfunction
 
 function! test#go#gotest#build_position(type, position) abort
-  if a:type == 'nearest'
-    let name = s:nearest_test(a:position)
-    if !empty(name) | let name = '-run '.shellescape(name, 1).' ./'.expand('%:h') | endif
-    return [name]
-  elseif a:type == 'file'
-    let path = fnamemodify(a:position['file'], ':h')
-    return path != '.' ? ['./' . path . '/...'] : []
-  else
+  if a:type == 'suite'
     return ['./...']
+  else
+    let path = './'.fnamemodify(a:position['file'], ':h')
+
+    if a:type == 'file'
+      return path == './.' ? [] : [path . '/...']
+    elseif a:type == 'nearest'
+      return s:build_nearest_args(a:position, path)
+    endif
   endif
 endfunction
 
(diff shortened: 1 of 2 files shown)


codeinabox:
Should this behaviour also be applied to `TestFile`?

gylaz:
This is already applied to `TestFile` (https://github.com/janko-m/vim-test/blob/master/spec/gotest_spec.vim#L35-L40).

janko-m:
Thank you for the PR! From what I understood, both `:TestNearest` and `:TestFile` should have the same "base" – the subdirectory of the package. So I'm wondering if this can be rewritten so that the same code is used for both. For example, your current code uses `%` for current file, however vim-test also supports running tests when current buffer is a non-test file, so it's better to use `a:position['file']` instead, like the `:TestFile` code uses.

Also, since I don't know Go, what is exactly the difference between `./mypackage` (used in `:TestNearest`) and `./maypackage/...` (used in `:TestFile`)? Meaning which one should be used for both? From what I understood, the `...` is for recursing into subdirectories, so maybe that should be used for nearest tests as well, but I'm not sure.

gylaz:
> From what I understood, both :TestNearest and :TestFile should have the same "base" – the subdirectory of the package. So I'm wondering if this can be rewritten so that the same code is used for both. For example, your current code uses % for current file, however vim-test also supports running tests when current buffer is a non-test file, so it's better to use a:position['file'] instead, like the :TestFile code uses.

Makes sense.

> From what I understood, the `...` is for recursing into subdirectories, so maybe that should be used for nearest tests as well, but I'm not sure.

Your understanding is correct, as long as subdirectories contain packages. 
From `go help packages`:
> Such a pattern expands to all package
directories found in the GOPATH trees with names matching the
patterns.  As a special case, x/... matches x as well as x's subdirectories.
For example, net/... expands to net and packages in its subdirectories.

I don't think we should use it for nearest tests. Say we run `go test -run 'TestFoo' ./...`. That command would find all the tests that matched `TestFoo` from the current package all the packages in subdirectories. If we want to be specific and ensure only the targeted test is ran, then I think `...` is not the right tool.

I think there is actually a bug in the current `TestFile` implementation. It will not only run the current test file, but all of the test files in the current package (and packages in subdirectories). I can open a separate PR to fix that.

gylaz:
Changes in 40e51d22d158e2d1ff4bd40c33475f64d217043f.

janko-m:
Thank you for your contribution!

PR #114, opened 2016-05-03
Provide filename for Go's TestFile and TestNearest

gylaz:
Providing a filename to the `go test` command ensures that only the
tests in that file are ran. This way if a similar test name exists in
another file (in the same package) it doesn't run.

### DIFF
--- a/autoload/test/go/gotest.vim
+++ b/autoload/test/go/gotest.vim
@@ -10,10 +10,10 @@ function! test#go#gotest#build_position(type, position) abort
   if a:type == 'suite'
     return ['./...']
   else
-    let path = './'.fnamemodify(a:position['file'], ':h')
+    let path = fnamemodify(a:position['file'], '.')
 
     if a:type == 'file'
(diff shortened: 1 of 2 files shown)


janko-m:
Thank you for the patch. It would be great if Go runner was that simple, but it seems that [this will work only in certain cases](http://stackoverflow.com/questions/16935965/how-to-run-test-cases-in-a-specified-file/16936314#16936314). Also, see https://github.com/janko-m/vim-test/pull/17.

On the other hand, not being able to tell `go test` to just run tests for a single test file is horrible. I don't know what is the right solution here.

gylaz:
@janko-m You are correct. Thanks for pointing that situation out. Perhaps keeping things as-is (`go run ./subdir`), is the best solution in this case. Feel free to close.

janko-m:
Ok, as I'm pretty sure that `go test` doesn't have a generic way for running a single test file (judging from #17), I will close this.

Issue #97, opened 2016-03-29
Support for Javascript + "tape"


bugeats:
Tape is a minimalistic test runner for Javascript:

https://github.com/substack/tape

As you can see from the above link, it also supports many formatters/reporters. From what I understand of vim-test so far, it looks like you could take your pick from one of those as long as core support for `tape` is added.

janko-m:
This seems like a reasonably popular JS test runner, so it makes sense to include it in vim-test. The only non-trivial part is figuring out how to correctly support formatters, since each one seems to be used a bit differently. I will try to get on it tomorrow.

janko-m:
I haven't found time for integrating tape, but it would be great if someone from tape users could write an integration. I will close this then, as it is not a feature that I have time to implement.

bugeats:
>  it would be great if someone from tape users could write an integration

Isn't that the point of this ticket? I recommend you keep tickets like this open so that people that want to contribute can have a reference.

aaalsaleh:
It is fairly easy to implement tape support, and I've already did it. I even allowed a `#reporter` config to specify one of the custom reporters if needed.

However, it is impossible to differentiate between the different javascript frameworks using just the `#test_file` methods. Both Mocha and Tape can have js files under 'test/tests' directory, and the current design of a mere file pattern check is insufficient to differentiate between the two. I had to check whether `node_modules/.bin/mocah` vs `node_modules/.bin/tape` existed along with the file pattern in order to differentiate between which one to use. For example:
```viml
function! test#javascript#mocha#test_file(file) abort
  return filereadable('node_modules/.bin/mocha') && a:file =~# g:test#javascript#mocha#file_pattern
endfunction
```
and
```viml
function! test#javascript#tape#test_file(file) abort
  return filereadable('node_modules/.bin/tape') && a:file =~# g:test#javascript#tape#file_pattern
endfunction
```
Checking these local binaries is the only way to differentiate between them, otherwise one of them will always take precedence.

janko-m:
Thank you for the investigation. Yes, that's why I made `#test_file` functions, so that in addition to pattern matching you can also inspect the environment for whatever you can find.

I like most your first proposed apporach, where you check for the executables in node_modules. It's great that you also thought of reporters, do you think that you could make the `#reporter` variable accept an array, which allows you to pipe multiple reporters at the end? I saw somewhere that this can be done and is useful. In that case it should probably be named `#reporters`.

aaalsaleh:
[...]

PR #119, opened 2016-06-03
Add support for JavaScript TAP-based testing frameworks

aaalsaleh:
Add support for JavaScript TAP-based testing frameworks, as requested in #97.
- It can detect a locally installed `tape` or `tap` (under the local `node_modules/.bin/*` only).
- It can be extended to support any other TAP-based frameworks using the config `g:test#javascript#tap#runners`.
- It can be configured to use any locally installed reporters through the config `g:test#javascript#tap#reporters`
- If multiple reporters are specified and only one of them is installed, it will be used as the reporter.
- If multiple reporters are specified and installed locally, they will be piped one after the other. Of course, all reporters except the last one must provide TAP output for the following reporter. Only certian reporters produce TAP output (e.g. `tap-notify`).
- If multiple reporters are specified and none of them is installed, no reporter will be used (default TAP output).
- Only `TestFile` and `TestSuite` are supported. Since there is no way to specify a certain test, `TestNearest` is the same as `TestFile`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,9 +7,9 @@ A Vim wrapper for running tests on different granularities.
 Currently the following testing frameworks are supported:
 
 | Language       | Frameworks                            | Identifiers                                  |
-| :------------: | ------------------------------------- | -------------------------------              |
+| :------------: | ------------------------------------- | -------------------------------------------- |
 | **Ruby**       | RSpec, [Minitest][minitest], Cucumber | `rspec`, `minitest`, `cucumber`              |
-| **JavaScript** | Intern, Mocha, Jasmine                | `intern`, `mocha`, `jasmine`                 |
+| **JavaScript** | Intern, TAP, Mocha, Jasmine           | `intern`, `tap`, `mocha`, `jasmine`          |
 | **Python**     | Nose, PyTest, Django                  | `nose`, `pytest`, `djangotest`, `djangonose` |
 | **Elixir**     | ExUnit, ESpec                         | `exunit`, `espec`                            |
 | **Go**         | Go                                    | `gotest`                                     |
--- a/autoload/test/javascript/tap.vim
+++ b/autoload/test/javascript/tap.vim
@@ -1,0 +1,44 @@
+if !exists('g:test#javascript#tap#file_pattern')
+  let g:test#javascript#tap#file_pattern = '\vtests?/.*\.js$'
+endif
+
+if !exists('g:test#javascript#tap#runners')
+  let g:test#javascript#tap#runners = ['tape', 'tap']
+endif
+
+if !exists('g:test#javascript#tap#reporters')
+  let g:test#javascript#tap#reporters = []
+endif
+
+function! test#javascript#tap#test_file(file) abort
+  return !empty(test#javascript#tap#executable()) && a:file =~# g:test#javascript#tap#file_pattern
+endfunction
+
+function! test#javascript#tap#build_position(type, position) abort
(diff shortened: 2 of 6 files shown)


janko-m:
Thank you so much! ❤️

I just have a few comments:

1. I like the generality of TAP-based testing frameworks that you've enabled. However, are there currently any such frameworks besides tape? Also, is the `tap` executable a convention for running TAP-based frameworks? Because although `tap-*` reporters will work for any TAP output, as far as I can tell they all only mention tape. So, even though the tape test.vim runner can be generalized, I'm more inclined to keep the name of the runner "tape" rather than "tap", since later on we might want to add tape-specific support (if tape introduces some options).

2. I don't think that `g:test#javascript#tap#runners` is necessary, since test.vim always allows you to change the executable for any runner by setting `g:test#{language}#{runner}#executable`. If in 1. we decide to keep only `tape`, maybe `test#javascript#tap#executable()` could just be:

  ```rb
  function! test#javascript#tap#executable() abort
    return 'node_modules/.bin/tape'
  endfunction
  ```

3. Judging from the tape README, I think that test suite pattern should be changed from `test/*.js` to `test/**/*.js` in order to properly recurse.

4. Great work with the reporters.

aaalsaleh:
Glad I could help.

1. Yes, there is also [tap](https://github.com/tapjs/node-tap), also some other reporters have their own dedicated cli, and there might be other testing framework that could output TAP stream now or in the future. No, `tap` is not the standard executable, as you can have any framework produce TAP stream. Actually, The [TAP](https://testanything.org) is a standard for output stream, and it is supported by multiple frameworks across multiple languages.

2. The good thing with having multiple runners and reporters is that (1) any runner found could be used with any reporter found removing the need to support each one, and (2) you can have different project using different TAP-based runners without needing to change the global `#executable` back and forth (e.g. between `tape` and `tap`).

3. This one is tricky. The `tap` support the `test/*` to run all tests and subtests, while `tape` introduced the glob `test/**/*.js` and doesn't support the former. Most importantly, the `test/**/*.js` assumes all tests are in sub-directories, and **those in the root `test/` will not be tested**. It will be impossible to test flat tests structure with `TestSuite` if it was changed to `test/**/*.js`. The common ground is `test/*.js` which is supported by both (but assume all tests in root `test/`). Regardless, I personally believe that `tape` should support `test/*` or `test/` which runs everything recursively.

4. Thanks!

I intentionally named it `tap` since I expected such comments. I think best action is either:

1. Have all TAP-based framework in a single `tap.vim`, and handle each specific framework option manually by checking which executable have been detected. Less redundancy, multiple simultaneous  support, and partial future support. However, more cases when handling special options (should be one `if-else` both returning array currently).

2. Keep the current `tap.vim` as it is, and migrate support for `tape` in a new `tape.vim`. Clear separation between `tape` and other TAP-based frameworks. However, more redundant code (almost the same), and more configurations (reporters).

I personally prefer the first, but wouldn't mind the second. Either case, I recommend adding TAP support for other languages since they share the same code (minus the executable).

janko-m:
Thank you very much for the detailed explanation, I now completely understand the reasoning for your implementation. I will merging shortly, and will then change the suite to `"test/**/*.js"` (with quoting).

On a side note, I don't understand why are there two test frameworks named by the *output* they generate. As I understood TAP is a general output, which means that you should theoretically be able to configure any test runner of any language to output in the TAP format. I mean, they're probably both great, it's just that for me the "framework" and "output" are two completely separate things.

PR #120, opened 2016-06-07
Update tslime link in README

jgeiger:
The previous link no longer exists and this seems to be the currently maintained fork.


janko-m:
Thanks!

PR #124, opened 2016-06-15
Add basic karma runner

### LINKED ISSUE #123: Support karma test runner
While `vim-test` supports the test suites `jasmine` and `mocha`, it does not support running tests through `karma`, which is quite common especially on angular projects. A new runner needs to be added to accomodate this configuration.

Kjir:
Add a basic karma runner which doesn't restrict the tests being run in any way due to the limitations in how `karma` selects the tests to run.

resolves #123

### DIFF
--- a/autoload/test/javascript/karma.vim
+++ b/autoload/test/javascript/karma.vim
@@ -1,0 +1,37 @@
+if !exists('g:test#javascript#karma#file_pattern')
+  let g:test#javascript#karma#file_pattern = '\v^spec/.*spec\.(js|jsx|coffee)$'
+endif
+
+function! test#javascript#karma#test_file(file) abort
+  return a:file =~? g:test#javascript#karma#file_pattern
+endfunction
+
+function! test#javascript#karma#build_position(type, position) abort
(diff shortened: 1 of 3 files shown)


janko-m:
Thanks a lot for the PR! ❤️

Kjir:
It's not perfect, I have in mind future improvements, whenever I manage to find the time I will make another PR 😄 

In case you're wondering the way we detect the tool to use should be improved since right now it overlaps with jasmine and it needs a custom configuration to be able to choose karma. That's against the projects philosophy so it needs to change!

janko-m:
@Kjir Ah, I didn't notice that. Maybe detecting `node_modules/.bin/karma` is enough, like in the [TAP runner](https://github.com/janko-m/vim-test/blob/ee4099c252a27da232815544da2832ae1d9a8580/autoload/test/javascript/tap.vim#L14)? In that case it should be added to Jasmine as well.

Kjir:
Yeah that's what I wanted to do! Thanks for pointing me to an appropriate example, I foresee some copy&paste ahead 😄

PR #126, opened 2016-06-24
Add runner for M, a Test::Unit runner that runs tests by line number

ruuts:
https://github.com/qrush/m

The default minitest runner sometimes has issues with the nested spec
syntax. Describe > describe etc.

Let's not reinvent the wheel, let's use M.

Example .vimrc settings:

" register the M runner.
let test#runners = {'Ruby': ['M']}
" you want to set the runner_commands because otherwise the default
" minitest runner will be used. In our case we had multiple test
" suites.. legacy!
let g:test#runner_commands = ['M', 'Cucumber', 'Rspec']

### DIFF
--- a/autoload/test/ruby/m.vim
+++ b/autoload/test/ruby/m.vim
@@ -1,0 +1,33 @@
+if !exists('g:test#ruby#m#file_pattern')
+  let g:test#ruby#m#file_pattern = '_test\.rb$'
+endif
+
+function! test#ruby#m#test_file(file) abort
+  return a:file =~# g:test#ruby#m#file_pattern
+endfunction
+
+function! test#ruby#m#build_position(type, position) abort
+  if a:type == 'nearest'
+    return [a:position['file'].':'.a:position['line']]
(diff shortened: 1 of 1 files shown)


janko-m:
Thanks for the pull request! Yeah, it's much better to have the test runner itself able to execute by line number, I didn't know about `m` or `maxitest` when I was created vim-test. I will merge shortly.

ruuts:
I didn't new about maxitest. Thanks! I think prefer M in our case since it's not changing minitest itself, it just sticks to being a runner.

janko-m:
In order to give people choice whether they want to use the M runner or Minitest's default runner (for reasons described in https://github.com/janko-m/vim-test/commit/153b5de3338c372fe9dd7138edb383013ddd5e84), I've made using M opt-in:

```vim
let g:test#ruby#minitest#executable = 'm'
```

ruuts:
@janko-m perfect! btw thanks for making vim-test. Really saves me a lot of time!

PR #128, opened 2016-06-26
Improve karma detection and add customizable karma.conf.js

### LINKED ISSUE #127: Improve karma test detection
Right now, the karma runner matches the same files as jasmine. The implementation should be improved to actually check for the correct binary. As suggested by @janko-m the [TAP runner](https://github.com/janko-m/vim-test/blob/ee4099c252a27da232815544da2832ae1d9a8580/autoload/test/javascript/tap.vim#L14) already lays the groundwork for this.

Kjir:
Improve the way we detect karma as our test runner so that it is more
relaxed on the file names, but will check for the karma executable and
the presence of the karma.conf.js file. The name and/or location of the
karma.conf.js can be customized by setting
g:test#javascript#karma#conffile.

Closes #127

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ Currently the following testing frameworks are supported:
 | Language       | Frameworks                                 | Identifiers                                  |
 | :------------: | -------------------------------------      | -------------------------------------------- |
 | **Ruby**       | RSpec, [Minitest][minitest], [M], Cucumber | `rspec`, `minitest`, `m`, `cucumber`         |
-| **JavaScript** | Intern, TAP, Mocha, Jasmine, Karma         | `intern`, `tap`, `mocha`, `jasmine`, `karma` |
+| **JavaScript** | Intern, TAP, Karma, Mocha, Jasmine         | `intern`, `tap`, `karma`, `mocha`, `jasmine` |
 | **Python**     | Nose, PyTest, Django                       | `nose`, `pytest`, `djangotest`, `djangonose` |
 | **Elixir**     | ExUnit, ESpec                              | `exunit`, `espec`                            |
 | **Go**         | Go                                         | `gotest`                                     |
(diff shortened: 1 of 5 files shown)


janko-m:
Thank you very much for the PR, this is a big improvement over the current state.

Firstly, I don't think that we have to define the `g:test#javascript#karma#executable` explicitly, since it is actually something that test.vim does automatically for every runner: if the user defines a `g:test#{language}#{runner}#executable` variable, test.vim will use that over `test#{language}#{runner}#executable()` function.

Secondly, in `test#javascript#karma#test_file()` is it maybe enough just to check for the existence of the executable and file pattern? Also, related to that, is it necessary to keep track of `karma.conf.js` in test.vim? Karma will automatically detect the following locations:

* `./karma.conf.js`
* `./karma.conf.coffee`
* `./.config/karma.conf.js`
* `./.config/karma.conf.coffee`

So I was thinking if the user really wants to put the configuration file somewhere other than these locations, they can just configure a generic CLI option in test.vim, which will automatically get appended to the test command:

```vim
let g:test#javascript#karma#options = 'path/to/karma.conf.js'
```

janko-m:
Also, I'm thinking of adding the `node_modules/.bin/{mocha,jasmine,karma,..}` executable check to every JavaScript runner on top of the file pattern, just like you did here, so that runners don't need to be declared in a specific order.

Kjir:
Thanks for the feedback, it's very useful!
About the karma config file: the check is there because if there is no karma conf file, the karma runner wouldn't work at all. On the other hand, if we find the karma executable in `node_modules/.bin/karma` we can presume that the current project uses karma and has a configuration file. The problem would only be if the user has set the executable to a karma binary in a shared location, in which case vim-test would most probably always choose karma whether a karma conf is present or not...

Kjir:
@janko-m I updated the commit following your advice! Hopefully this is even cleaner that before...

janko-m:
This looks great now, thank you very much! ❤️

PR #131, opened 2016-07-13
Updated Karma Runner

trisys3:
This is my first pull request, so apologies if I did anything wrong. I tried to follow the general style of the rest of the repo, but I had to deviate in certain ways, as I detail below.

Basically, I added the ability to pass options to karma through an intermediary JavaScript file. This is because karma's CLI isn't all that flexible, because they mainly focus on the configuration file, karma.conf.js. Mainly, it doesn't work too well with options that are supposed to be arrays. On my project, I followed the example of the gulp-karma and grunt-karma plugins, which added an intermediary file parsing out the arguments & passing them to karma. I just used what I learned there on this repo.

The intermediary file I added does not have precedence in this repo, but syntastic has some examples. See [the less syntax-checker folder](https://github.com/scrooloose/syntastic/tree/master/syntax_checkers/less), where they add a CoffeeScript file and compile it to JavaScript before committing.

If you can't accept my pull request for whatever reason, I understand, but please consider it. Thanks.


trisys3:
Sorry about the conflicts, I didn't notice them before I created the pull requests. I'm still new to them. Should I resolve the conflicts and then request again, or would someone need to delete this request first?

trisys3:
Resolved the conflicts.

trisys3:
The jasmine tests are failing. They were failing when I merged, before I resolved the conflicts, but I didn't check out a brand-new branch to check if this is happening on the current master.

trisys3:
I can confirm that this is only happening on my branch, and I am investigating.

trisys3:
The merge should be fine now. I hope you like it.

janko-m:
Thank you for your work of improving Karma granularity, I really appreciate it! ❤️ 

It's a bit unfortunate that Karma, even though it is very configurable by code, it doesn't have a runner which supports this configuration via CLI options. But it's awesome that you made one such runner, and I definitely want to use it in test.vim. Do you think it makes sense to release this as a Node module? I think it would greatly benefit any Karma user, regardless of whether they're using test.vim, and that way it can be maintained and improved separately. It seems much more advanced than grunt-karma and gulp-karma, and the advantage is that people can use it for every environment.

Then we could make test.vim automatically detect whether your module is installed, and use all of its benefits, otherwise falling back to regular `karma`. We would also make add additional test.vim documentation noting this benefit. What do you think?

trisys3:
Oh thanks, it was nothing. You're right, it is rather unfortunate that `karma` doesn't put more effort into its own runner, but then again, we are talking about an open-source project led by a couple people at a multi-million dollar company in their spare time. We can cut them some slack. Plus, the only real problem with the CLI options, besides the merging with `karma.conf.js`, seems to be values that are supposed to be arrays. Unfortunately, this includes the `--files` option.

Regarding the runner I made ("hacked up" might be a better description), I'm sure there are better options out there, but I can't seem to find any at the moment. As I said, I already made a slightly more specialized version for use at my organization, so it was pretty easy, all I had to do was remember what I did before. If you think it would be beneficial for me to release it on my own, I'm all for it. Until I can get it released, though, it should probably live here.

Next, I'll try my hand at `protractor`, which I noticed is missing from your stable of testers.

trisys3:
OK, I did as you suggested and published my args parser as an npm package. It is really simple right now, as indicated by the 0.0.1 version, but it works, and I will improve it when I have time again.

janko-m:
Great job with publishing your better CLI runner for Karma, and sorry for taking time to merge this, I'm just busy lately.

Just wanted to check, does your runner also recognize regular Karma options like `--single-run`, `--no-auto-watch` and `--log-level=OFF`? I mean, I suppose it does if you added them, just want to check.

After that this PR is ready for merge!

trisys3:
I understand, I've been busy, too, during the week, and on the weekend sometimes it's hard to get myself to do things that resemble what I did all week.

Yes, my runner, [karma-cli-runner](https://www.npmjs.com/package/karma-cli-runner), accepts all these options. I originally had them as defaults in my parser, when I first broke it apart from `vim-test`, then I realized it's probably better to keep the API as close as possible to `karma`'s. I'm still not sure what to do about `karma start`/`karma run`. I'm tempted to just support `start` and not worry about `run`, especially since `run` needs a server to connect to, meaning you already need to have done `start`.

My runner is really simple, all it does is parse all the options and pass them to `karma`. There are already parsers that do this, and `karma` uses one itself. I just wanted to have as few dependencies as possible, for cases such as yours, where dependencies are hard to deal with, especially considering `karma` is already a dependency. I also wasn't sure how to handle `node` modules in your plugin, as `vim-flavor` doesn't seem to install plugins that aren't `vim` plugins and I didn't want to bog down your plugin with deps that are only used in tests. If there is some sort of post-install hook that installs deps of other languages, that might be better, for `karma` as well as `jasmine`, etc., as then these would not be locked to the same version in tests, among other things.

trisys3:
By the way, sorry, but I'm not going to add `protractor`. I just remembered it's even harder to deal with than `karma`, and unlike `karma`, has no API, you essentially have to run it as an executable. It sort of worked for my organization because we have a special set of needs, but generalizing it would be a nightmare.

trisys3:
Is there anything preventing this pull request? Did I do anything wrong? Otherwise, could it be merged, please?

janko-m:
@trisys3 There were some minor things that I wanted to add but didn't want to bother you with them, and never got to finalizing it. Thank you for pushing me, and sorry for the delay 😃

trisys3:
No problem, just want to help!

PR #132, opened 2016-07-28
Install instructions are for vim-plug

Insti:
Clarify that the install instructions are for vim-plug.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,9 +38,11 @@ in, so they all work in the same unified way.
 
 ## Setup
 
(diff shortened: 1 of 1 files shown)


janko-m:
Thanks! I original wrote it like `Plug[in]` so that it signals both vim-plug's `Plug` and Vundle.vim's `Plugin`, but I agree that might look confusing, and also it doesn't cover Pathogen.vim. And I think that there is no reason to ever use another plugin manager than vim-plug, so I like this change.

PR #134, opened 2016-08-04
async tests support for python

purpleP:
In python 3.5 syntax for creating coroutines on a language level have been added.
Because of that libraries like pytest-asyncio have been created, that can decorate test that is written as a coroutine object and run it in a loop.

```
@pytest.mark.asyncio
async def test_some():
    assert await some_awaitable() == some_expected_result


janko-m:
Looks good, thank you!

purpleP:
@janko-m You're welcome.

PR #137, opened 2016-08-13
Fix failing tests for Elixir umbrella applications

### LINKED ISSUE #136: Elixir umbrella support
Hi,

I tried to use your plugin to run tests inside an Elixir umbrella app but it doesn't seem to work.

It looks like `vim-test` run its test command using a relative path instead of an absolute path.  

Relative path:

```
mix test apps/app_name/test/models/model_name_test.exs
# ==> app1
# Test patterns did not match any file: apps/app1/test/file_test.exs
# ==> app2
# Test patterns did not match any file: apps/app1/test/file_test.exs
```

Absolute path:

```
mix test /Users/kdisneur/my_umbrella/apps/my_app/test/models/model_name_test.exs`
# ==> app1
# ....
#
# Finished in 0.1 seconds
#4 tests, 0 failures
#
# Randomized with seed 549588
# ==> app2
#
#
# Finished in 0.00 seconds
#0 failures
```

I also tried to run the command `mix test` using an absolute path, on a "normal" project. And it's working properly. 

Do you think it would be possible to use absolute path instead of relative path for Elixir tests ?

Thanks

kdisneur:
When running a specific test in Elixir, the file has to be either an absolute path or a path relative to the umbrella app.

As it's not easily possible to know if a project is under an umbrella application, the fix is to use an absolute path.

Fixes #136

### DIFF
--- a/autoload/test/elixir/exunit.vim
+++ b/autoload/test/elixir/exunit.vim
@@ -7,17 +7,19 @@ function! test#elixir#exunit#test_file(file) abort
 endfunction
 
 function! test#elixir#exunit#build_position(type, position) abort
+  let file = fnamemodify(a:position['file'], ':p')
+
   if test#elixir#exunit#executable() == 'mix test'
     if a:type == 'nearest'
-      return [a:position['file'].':'.a:position['line']]
+      return [file.':'.a:position['line']]
     elseif a:type == 'file'
-      return [a:position['file']]
+      return [file]
     else
       return []
     endif
   else
     if a:type == 'nearest' || a:type == 'file'
-      return [a:position['file']]
+      return [file]
     else
       return ['*.exs']
     end
--- a/spec/exunit_spec.vim
+++ b/spec/exunit_spec.vim
@@ -16,14 +16,14 @@ describe "ExUnit"
       view +1 normal_test.exs
       TestNearest
 
-      Expect g:test#last_command == 'mix test normal_test.exs:1'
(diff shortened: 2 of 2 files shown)


janko-m:
Thanks for the patch!

However, there already is a generic way to use absolute file paths in test.vim, by changing the [filename modifier](https://github.com/janko-m/vim-test#filename-modifier):

```vim
let test#filename_modifier = ':p'
```

I think it's good that it's opt-in, because someone who doesn't need absolute paths can still choose to keep relative paths for readability.

kdisneur:
Cool - I missed this option in the documentation.

I tried it locally but I'm not able to opt-in for a specific programming language. Is it a global option only?

If yes, do you think we could make it customisable per programming language? If it's not possible do you think we could add an option specifically for Elixir? 

To be honest, I dont feel really comfortable to activate "absolute path" globally when only one kind of project need it. :/ 

The absolute path seems to be required for any umbrella project (http://elixir-lang.org/getting-started/mix-otp/dependencies-and-umbrella-apps.html#umbrella-projects), it's not specific to my own setup. Meaning any dev who wants to create an umbrella app in Elixir will need to switch to absolute path.

janko-m:
You could set the filename modifier in a `.vimrc` located in your project directory, and then it will be applied only when you open a Vim session in that project. You just need to `set exrc` in your global `~/.vimrc`.

Alternatively, in ExUnit/ESpec runners we could try detecting whether the current project is an umbrella project, and then set the filename modifier to absolute.

PR #140, opened 2016-08-17
Fix incorrect log level for karma

wilcus:
According to http://karma-runner.github.io/1.0/config/configuration-file.html


janko-m:
Thanks for the fix!

PR #142, opened 2016-08-25
Fix 'basic' startegy on Windows

BimbaLaszlo:
When I like to run a test, the `cmd.exe` says:

  The system cannot find the path specified

The source of this problem is that Windows using `&` to separate
commands, `;` for command line parameters.

See:
www.microsoft.com/resources/documentation/windows/xp/all/proddocs/en-us/ntcmds_shelloverview.mspx


janko-m:
Thanks for the patch!

PR #143, opened 2016-08-25
Add VimProc as a new strategy

BimbaLaszlo:
To catch the results of the test to a buffer, use `:Verbose TestFile` with Tim Pope's Scriptease plugin.


janko-m:
Thank you!

PR #148, opened 2016-10-18
Added Javascript Ava runner

edelprino:
I tried to add Ava (https://github.com/avajs/ava) Javascript test runner based on mocha runner.

Unfortunately, now, there is the same problem with python. There are multiple test runner with the same file pattern. I added, like python, a variable `g:test#javascript#runner` that you can choose witch runner use.

I've also update the `Readme.md` with specific instruction for Javascript.

What's do you think?
Thanks

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,21 +6,21 @@ A Vim wrapper for running tests on different granularities.
 
 Currently the following testing frameworks are supported:
 
-| Language       | Frameworks                                 | Identifiers                                  |
-| :------------: | -------------------------------------      | -------------------------------------------- |
-| **Ruby**       | RSpec, [Minitest][minitest], [M], Cucumber | `rspec`, `minitest`, `m`, `cucumber`         |
-| **JavaScript** | Intern, TAP, Karma, Mocha, Jasmine         | `intern`, `tap`, `karma`, `mocha`, `jasmine` |
-| **Python**     | Nose, PyTest, Django                       | `nose`, `pytest`, `djangotest`, `djangonose` |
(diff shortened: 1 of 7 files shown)


janko-m:
Thanks for the pull request! ❤️

I think you can detect whether the project is using the AVA runner by searching for the `"ava"` string in `package.json`.

edelprino:
You're welcome!
Maybe if you install `ava`globally your `package.json` doesn't contains `"ava"`string. Search in `package.json` maybe can be little bit "magic".

I don't know, for me work well, but I don't know if is better leave this choice explicit and leave the global variable or put a check on `package.json`.

codeinabox:
@edelprino where would the ava binary been installed if it was part of the package.json? I found detecting the binary the simpler approach which I used in #178

sloria:
I am currently working on a [Lab](https://github.com/hapijs/lab) runner, and I'm running into the same problem as this PR with having multiple runners for the same file pattern. +1 to adding the ` g:test#javascript#runner` option. It's nice to be able to specify your preferred runner explicitly.

codeinabox:
@edelprino do you need any further help with this PR? We've now added support for package.json detection of test runners which would help unblock this.

tylucaskelley:
Any update on this? Would love to see ava support :)

codeinabox:
I am happy to help @edelprino with this branch. Otherwise I can pick it up after Christmas.

codeinabox:
Closing this PR as the changes have been moved into #258

PR #152, opened 2016-11-25
test_nearest is executing exactly one test

purpleP:
(pytest test_some.py::test_nearest vs pytest test_some.py -k test_nearest)


janko-m:
@purpleP Thanks!

PR #154, opened 2016-11-27
Add Codeception support (PHP test framework)

pbogut:
I've added support for [Codeception](http://codeception.com/).

At current state, it support `*Cest.php`, `*Cept.php` and `*.feature`. 
The one issue I've got is that Codeception supports `*Test.php` as well, but these are highjacked by `PHPUnit`.  Is there any solution for that? I'm checking for Codeception specific files, but PHPUnit takes precedence. 

I'm not sure if I can change check order, or there is another way?


pbogut:
I figured out I can implement `g:test#php#runner` in a similar way how it is done with `pytest` and `nose`.
@janko-m what's your thoughts?

janko-m:
First of all, thank you very much for the PR! :heart:

For detecting Codeception, many vim-test runners do additional detection in the `test_file()` function, something in the project repository that indicates the user is using Codeception. For example, for JavaScript testing frameworks we would read `package.json`.

Is there something with which we can detect that Codeception is being used? In that case we should also move Codeception in the beginning of the list. Otherwise your solution is good, as it is the only way.

pbogut:
It could be checked in `composer.json`, but one can use it globally or as a phar package and that won't be in `composer.json`.

Another solution could be checking against  file structure. There is always `tests/_support/_generated` folder present, which seems to be fairly unique.

Do you want me to implement it based on folder structure and move Codeception before PHPUnit?

pbogut:
I don't know how could I missed that, but there is `codeception.yml` configuration file present in a root project directory. I can change it to check against that.

janko-m:
@pbogut That sounds perfect 👍

pbogut:
@janko-m It's done. Don't know why there is conflict in `README.md`, though.

janko-m:
@pbogut It's because you need to *rebase* on or *merge* master, instead of adding the same change that's in master.

pbogut:
I did rebase on your master. Is that all right?

pbogut:
@janko-m, how you feel about merging it?

janko-m:
@pbogut Sorry for the delay, thanks for reminding me.

It looks that this isn't properly rebased/merged, because we can see the commits from master here, and we can also see them in the diff. I cherry-picked 2c530af which is the one that adds Codeception support.

Thank you so much for you work on this, now test.vim has another test runner! ❤️

pbogut:
Sorry about that, after rebase I couldn't push without merging back and then, well, you have seen the diff. Thank you for merging I can use master now :)

PR #160, opened 2016-12-19
Change rails version detection method

5t111111:
I am not exactly sure why though, at least for my environment, Rails version detection always fails because `cat Gemfile.lock` returns long one-liner which includes newline symbols and the regex also fails because of leading whitespaces in `Gemfile.lock`.

So I have changed the Rails version detection method to more reliable one using Vim builtin function.

How does this sound to you?

FYI: 
macOS 10.12.1

```
$ vim --version                                                                                                                                                    
VIM - Vi IMproved 8.0 (2016 Sep 12, compiled Dec 16 2016 10:35:52)
MacOS X (unix) version
Included patches: 1-134
Compiled by Homebrew
```

### DIFF
--- a/autoload/test/ruby/rails.vim
+++ b/autoload/test/ruby/rails.vim
@@ -38,10 +38,14 @@ endfunction
 
 function! s:rails_version()
   if filereadable('Gemfile.lock')
-    let gemfile_lock = system('cat Gemfile.lock')
-    let version_string = matchstr(gemfile_lock, '\v^rails \(\zs.+\ze\)')
+    for line in readfile('Gemfile.lock')
+      let version_string = matchstr(line, '\v^ *rails \(\zs\d+\.\d+\..+\ze\)')
+      if version_string
+        break
+      endif
+    endfor
 
-    if !empty(version_string)
+    if version_string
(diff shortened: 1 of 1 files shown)


janko-m:
@5t111111 Thanks for the patch! This looks good, I'm just curious why the change from `rails \(\zs.+\ze\)` to `rails \(\zs\d+\.\d+\..+\ze\)`? Is there some scenario where the former regex wouldn't work?

Btw, TIL about `readfile()`! It's much better not to have to rely on `cat`, which I'm not even sure exists on Windows.

5t111111:
> I'm just curious why the change from rails \(\zs.+\ze\) to rails \(\zs\d+\.\d+\..+\ze\)? 

The reason is the former regex also matches rails defined as dependency in `Gemfile.lock`, e.g.:

```
...

GEM
  remote: https://rubygems.org/
  specs:

    ...

    rails (5.0.0.1) #<= expect to match this line

...

    activeadmin (1.0.0.pre4)

      ...

      rails (>= 3.2, < 5.1) #<= it also matches this line (another gem depends on this version of Rails)

...

DEPENDENCIES

...

  rails (~> 5.0.0) #<= it also matches this line 
```

janko-m:
@5t111111 Ok, that makes total sense, I didn't think about it. Great, I'll merge shortly.

PR #161, opened 2016-12-20
nearest strategy now executes only one test and test classes work also

purpleP:
Thanks to github it's all in the name. I've found some time to properly implement this. I've tested two use cases. One when test is a module level function and one where test is a method of a class.
Wasn't very hard to do, because you've already searching for namespace and it's available. The only thing that stopped me before is that my knowledge of vimscript was worse than it is now.

### DIFF
--- a/autoload/test/python/pytest.vim
+++ b/autoload/test/python/pytest.vim
@@ -16,7 +16,7 @@ function! test#python#pytest#build_position(type, position) abort
   if a:type == 'nearest'
     let name = s:nearest_test(a:position)
     if !empty(name)
-      return [a:position['file'].' -k '.name]
+      return [a:position['file'].'::'.name]
     else
       return [a:position['file']]
     endif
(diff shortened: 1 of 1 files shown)


janko-m:
@purpleP Thank you very much! Looks good, I'll merge shortly.

janko-m:
@purpleP Note that I added https://github.com/janko-m/vim-test/commit/37d4130d4cb36ba6ca88b3a69327947db48212db which updates tests, and adds support for running tests for a class by having the cursor above it (in that case we would get `py.test test_something.py::TestSomething::`, so I just corrected it to be `py.test test_something.py::TestSomething`).

purpleP:
@janko-m Oh, I didn't know that this is possible. I don't use TestClasses. I should learn vimspec...

janko-m:
@purplP I think it's great that you're learning it on the way, I learned most of my vimscript just from writing test.vim 😉

PR #165, opened 2017-01-04
If there is a nwb config present then run `nwb test`

codeinabox:
I am currently working on application using [nwb which requires you to run nwb test](https://github.com/insin/nwb/blob/master/docs/Commands.md#test). This is a quick PR to get vim-test to check for the presence of the nwb config.


janko-m:
Thanks for that patch!

PR #167, opened 2017-01-15
Improve documentation about asyncrun strategy

okuramasafumi:
- Added link to asyncrun plugin to README
- Added note saying that asyncrun strategy requires AsyncRun plugin


janko-m:

PR #175, opened 2017-03-08
Add support for Jest

smarquez1:
Runs tests from the `__tests__` directory, file names must end in `-test.js|jsx|coffee` .

I can make the file pattern more flexible, that's open to discussion.


ianks:
This would be nice... Are there any blockers that need to be fixed @janko-m so we can get this in?

janko-m:
@smarquez1 Thank you so much for this contribution, and I really appreciate the extensive test coverage! ❤️

PR #178, opened 2017-03-24
Adds support for PHP test frameworks Kahlan and Peridot

codeinabox:
This adds basic support for PHP test frameworks Kahlan and Peridot. Handling test nearest is forthcoming.

### DIFF
--- a/autoload/test/php/phpspec.vim
+++ b/autoload/test/php/phpspec.vim
@@ -3,7 +3,9 @@ if !exists('g:test#php#phpspec#file_pattern')
 endif
 
 function! test#php#phpspec#test_file(file) abort
-  return a:file =~# g:test#php#phpspec#file_pattern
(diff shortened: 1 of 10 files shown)


autoload/test/php/phpspec.vim
```diff
@@ -3,7 +3,9 @@ if !exists('g:test#php#phpspec#file_pattern')
 endif
 
 function! test#php#phpspec#test_file(file) abort
-  return a:file =~# g:test#php#phpspec#file_pattern
+  if a:file =~# g:test#php#phpspec#file_pattern
+    return search("PhpSpec") > 0
```
codeinabox: I wanted to search for `PhpSpec\ObjectBehavior` but it doesn't seem to like slashes in a search query and I couldn't figure out how to escape the,

PR #179, opened 2017-04-04
Add dotnet test runner

markwoodhall:
This pull request adds a runner for the dotnet cli command `dotnet test`. 

I'd appreciate any input about the logic I am using to find the nearest `.csproj` file, I'm not great with vimscript. Finding the nearest `.csproj` file is required because the path is a required argument to `dotnet test`. I've tested this out using an underlying xunit test project as well as an mstest one, both work fine. All of the functionality `:TestFile`, `:TestSuite`, and `:TestNearest` work as expected.

### DIFF
--- a/autoload/test/cs/dotnettest.vim
+++ b/autoload/test/cs/dotnettest.vim
@@ -1,5 +1,5 @@
 if !exists('g:test#cs#dotnettest#file_pattern')
-  let g:test#cs#dotnettest#file_pattern = '\v^.*[Tt]ests\.cs$'
+  let g:test#cs#dotnettest#file_pattern = '\v^.*\.cs$'
 endif
 
 function! test#cs#dotnettest#test_file(file) abort
--- a/autoload/test/cs/dotnettest.vim
+++ b/autoload/test/cs/dotnettest.vim
@@ -1,0 +1,59 @@
+if !exists('g:test#cs#dotnettest#file_pattern')
+  let g:test#cs#dotnettest#file_pattern = '\v^.*[Tt]ests\.cs$'
+endif
+
+function! test#cs#dotnettest#test_file(file) abort
+  return a:file =~? g:test#cs#dotnettest#file_pattern
+endfunction
+
+function! test#cs#dotnettest#build_position(type, position) abort
+  let file = a:position['file']
+  let filename = fnamemodify(file, ':t:r')
+  let filedir = fnamemodify(file, ':p:h')
+  let projfiles = split(glob(filedir . '/*.csproj'), '\n')
+  let search_for_csproj = 1
+
+  while len(projfiles) == 0 && search_for_csproj
+    let filedirparts = split(filedir, '/') 
+    let search_for_csproj = len(filedir) > 1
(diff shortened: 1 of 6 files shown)


janko-m:
Thank you for the pull request! ❤️

> I'd appreciate any input about the logic I am using to find the nearest .csproj file, I'm not great with vimscript. Finding the nearest .csproj file is required because the path is a required argument to dotnet test.

Hmm, test.vim pretty much has a constraint that you have to run Vim from the project root, but I guess `.csproj` isn't located in the project root?

autoload/test/cs/dotnettest.vim
```diff
@@ -0,0 +1,59 @@
...
+  let filter = ''
+  if a:args[0] != '*'
+    let filter = ' --filter FullyQualifiedName\~'.a:args[0]
+  endif
+  let args = ['test ', a:args[1], filter]
+  return [join(args, "")]
+endfunction
+
+function! test#cs#dotnettest#executable() abort
+  return 'dotnet'
```
janko-m: I find it nicer if we `return "dotnet test"` here, and then exclude it from `build_args`.

markwoodhall:
Thanks for the comments. I will address the issues and add a few tests.  

Most of the time in .net land people will be working at the solution level, so there may be many csproj files as part of the solution and they will almost always been in nested sub directories so I couldn't see any option but to walk up the tree  and find the nearest.

markwoodhall:
I've tried to write some tests but I'm having problems running them. 

img

Not sure if vim-flavor works with neovim?

janko-m:
@markwoodhall I didn't try running `vim-flavor` with neovim yet, but if it really doesn't work, you can always use the CI (it's passing for your last commit).

markwoodhall:
I will install vim and try that first.

markwoodhall:
For info, it works fine with vim installed.

janko-m:
Looks great! 👏

Just one last question: I was just wondering what was the reason for [this change](https://github.com/janko-m/vim-test/pull/179/commits/93bb213b6e804e56e52952c897cc84a0663661fb) in the test file pattern? One advantage of having the test file pattern *not* match application files is that you can run `:TestFile`, `:TestNearest` and `:TestSuite` from the application files, and since test.vim would know these aren't test files it would instead run the test command on the last test file that you ran tests on. I mean, definitely isn't a crucial feature, I just personally prefer to use that over `:TestLast` since it's easier for me to have only 3 Vim commands automatized than 4.

markwoodhall:
I changed that because it was proving difficult to pick a pattern that would actually make it work for the wide variety of conventions people use when naming their tests. I guess people could override it to match their convention but it felt more user friendly to cover a wider range to start with.

janko-m:
> I changed that because it was proving difficult to pick a pattern that would actually make it work for the wide variety of conventions people use when naming their tests. I guess people could override it to match their convention but it felt more user friendly to cover a wider range to start with.

Sounds great to me, thanks for the explanation 👍

Excellent, this is ready for merge, thank you again 🎉

markwoodhall:
No problem, thanks for the help along the way. :+1:

PR #182, opened 2017-04-11
Update jest.vim

spheric:
Allow `__test__` to be in sub-directory and allow `-` or `_` test suffix.


janko-m:
@spheric Looks good, thanks!

PR #183, opened 2017-04-14
Create test#cd_before_run config

alexgenco:
This config value determines whether or not to cd into the project's
working directory before running the test. If not set, it will default
to true, which matches the existing behavior.

This could potentially lead to some confusion since, for example `test#strategy#basic` doesn't always do the `cd`, so this option would have no effect.

I think this is a good change because for example when using `rvm` sometimes a `cd` can have a lot of output and be slow.

### DIFF
--- a/README.md
+++ b/README.md
@@ -244,6 +244,17 @@ different working directory for running tests:
 let test#project_root = "/path/to/your/project"
 ```
 
+Similarly, some strategies will first issue a `cd` into vim's working directory
+before running. You can tell test.vim to disable this behavior:
+
+```vim
+let test#strategy#cd_before_run = 0
+```
+
+Note that if `test#project_root` is set, test.vim will still issue a `cd` into
+that directory before running, even if `test#strategy#cd_before_run` is
+disabled. It just won't perform any additional strategy-specific `cd`s.
+
 ### Language-specific
 
 #### Python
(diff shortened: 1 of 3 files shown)


alexgenco:
I didn't write any tests yet because I didn't see an obvious place to add it (although it should probably be somewhere in `strategy_spec.vim`). If you think this is a good idea I'll figure out a good place for the test.

alexgenco:
I updated the docs and realized the interactions between this option and `test#project_root` are a little confusing so I tried to clarify in the README. I also renamed it to `test#strategy#cd_before_run` to make it clearer that this only affects certain strategies.

codeinabox:
Is it possible to merely suppress that output instead of making the behaviour configurable? Strategies an the inner workings on vim-test are not my speciality

> I think this is a good change because for example when using rvm sometimes a cd can have a lot of output and be slow.

janko-m:
@alexgenco What is the use case for preventing test.vim to `cd` into project's working directory? I'm aware that this logic may not be perfect, and perhaps should even be the behaviour for the `basic` strategy, because there it technically doesn't have any effect (Vim `getcwd()` will always equal to Vim's project directory); it was only meant for remote strategies like Vimux, Terminal.app or Iterm.app. I just wanted to check if there isn't a way where test.vim can automatically "do the right thing", and then we can avoid adding a new configuration option.

alexgenco:
Our setup at work was causing this `cd` to add a good amount of overhead to running a test over vimux due to `rvm`. However, when I opened this PR, I thought the slowness was causing the tests not to run at all, but that turned out not to be the case. So while I generally like the idea of making behaviors like these configurable, I don't think this particular change is all that valuable. I'll close for now, feel free to re-open if you think it's salvageable.

PR #184, opened 2017-04-22
Add lab support

sloria:
Add support for the [Lab](https://github.com/hapijs/lab) test runner.

Also adds `g:test#javascript#runner` because multiple runners can
handle files in test/*.js. #148 ran into this as well.

### DIFF
--- a/autoload/test/javascript/lab.vim
+++ b/autoload/test/javascript/lab.vim
@@ -1,0 +1,46 @@
+if !exists('g:test#javascript#lab#file_pattern')
+  let g:test#javascript#lab#file_pattern = '\vtest/.*\.js$'
+endif
+
+function! test#javascript#lab#test_file(file) abort
+  if a:file =~# g:test#javascript#lab#file_pattern
+    if exists('g:test#javascript#runner')
+      return g:test#javascript#runner == 'lab'
+    else
+      return executable('lab')
+    endif
+  endif
+endfunction
+
+function! test#javascript#lab#build_position(type, position) abort
+  if a:type == 'nearest'
+    let name = s:nearest_test(a:position)
+    if !empty(name)
+      let name = '--grep '.shellescape(name, 1)
+    endif
+    return [a:position['file'], name]
+  elseif a:type == 'file'
+    return [a:position['file']]
+  else
+    return []
+  endif
+endfunction
+
+function! test#javascript#lab#build_args(args) abort
+  return a:args
+endfunction
+
+function! test#javascript#lab#executable() abort
+  if filereadable('node_modules/.bin/lab')
+    return 'node_modules/.bin/lab'
+  else
+    return 'lab'
+  endif
+endfunction
+
+function! s:nearest_test(position)
+  let name = test#base#nearest_test(a:position, g:test#javascript#patterns)
+  return (len(name['namespace']) ? '^' : '') .
+       \ test#base#escape_regex(join(name['namespace'] + name['test'])) .
+       \ (len(name['test']) ? '$' : '')
+endfunction
--- a/autoload/test/javascript/lab.vim
+++ b/autoload/test/javascript/lab.vim
@@ -4,11 +4,13 @@ endif
 
 function! test#javascript#lab#test_file(file) abort
   if a:file =~# g:test#javascript#lab#file_pattern
-    if exists('g:test#javascript#runner')
-      return g:test#javascript#runner == 'lab'
-    else
-      return executable('lab')
-    endif
+    for line in readfile(a:file)
+      let pattern = '\v[Ll]ab.script\(\)'
+      if line =~# pattern
+        return 1
+      endif
+    endfor
+    return 0
   endif
 endfunction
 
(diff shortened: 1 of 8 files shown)


sloria:
lab can either be installed global or declared as a dependency.

janko-m:
@sloria First of all, thank you for the pull request! ❤️

The reason why @codeinabox was asking about installed as a dependency vs globally is because, if `lab` is declared as a dependency in `package.json` a file matches the pattern, then test.vim can be positive that `lab` is the correct runner to use, and not for example `mocha` (unless the user for some reason declares both `lab` and `mocha` in `package.json`, which would be fairly uncommon, I think). What we're wondering is if it's common in Node projects *not* to declare your test runner in `package.json` and use the globally installed one (in which case we should probably go with `test#javascript#runner`), or is that uncommon/non-idiomatic, in which case checking presence of `lab` in `package.json` should be ok.

autoload/test/javascript/lab.vim
```diff
@@ -0,0 +1,50 @@
...
+function! test#javascript#lab#test_file(file) abort
+  return a:file =~# g:test#javascript#lab#file_pattern
+endfunction
+
+function! test#javascript#lab#test_file(file) abort
+  if a:file =~# g:test#javascript#lab#file_pattern
+    if exists('g:test#javascript#runner')
+      return g:test#javascript#runner == 'lab'
+    else
+      return executable('lab')
```
janko-m: Here we could rather check `filereadable('node_modules/.bin/lab')`, if we decide we want to make declaring `lab` as a dependency in `package.json` mandatory.

spec/fixtures/lab/outside.js
```diff
@@ -0,0 +1,5 @@
+describe('Addition', function() {
+  it('adds two numbers', function() {
+    assertEqual(1 + 1 == 2)
+  });
+});
```
janko-m: We don't need this file for tests, Vim can `view outside.js` even if the file doesn't exist – in that case it opens a new buffer.

autoload/test/javascript/lab.vim
codeinabox: Exactly, this is what I did for my PR https://github.com/janko-m/vim-test/pull/178 where the test file format was the same but the runner is determined by which one is installed in the project binary directory

codeinabox:
I think at this stage it would probably be easier to check which binary is present in the project's node modules directory as I can't imagine any project having multiple test runners in their dev dependencies. As I mentioned this is what I did for #178 

> The reason why @codeinabox was asking about installed as a dependency vs globally is because, if lab is declared as a dependency in package.json a file matches the pattern, then test.vim can be positive that lab is the correct runner to use, and not for example mocha (unless the user for some reason declares both lab and mocha in package.json, which would be fairly uncommon, I think). What we're wondering is if it's common in Node projects not to declare your test runner in package.json and use the globally installed one (in which case we should probably go with test#javascript#runner), or is that uncommon/non-idiomatic, in which case checking presence of lab in package.json should be ok.

sloria:
@janko-m @codeinabox Thank you for reviewing my PR. I understand why we would want to detect the presence of `lab` in dev dependencies, given vim-test's "Do The Right Thing" philosophy. I agree that it would be ideal to automatically choose the right runner.

Unfortunately, there doesn't seem to be community-wide consensus on whether to install test runners globally or declare them in dev dependencies. I've seen projects that do one and not the other.

That said, one way I might be able to work around this is to search the current test file for the text `Lab.script()`, which should be present in any lab test file. 

There would still be the issue of making `:TestSuite` work though, right? If that doesn't work even with the workaround, then we'd still need to add `test#javascript#runner`.

autoload/test/javascript/lab.vim
```diff
@@ -0,0 +1,50 @@
+if !exists('g:test#javascript#lab#file_pattern')
+  let g:test#javascript#lab#file_pattern = '\vtest/.*\.js$'
+endif
+
+function! test#javascript#lab#test_file(file) abort
+  return a:file =~# g:test#javascript#lab#file_pattern
+endfunction
```
sloria: Well that's an embarrassing oversight. Fixed.

janko-m:
@sloria `:TestSuite` detects which runner to run based on the last test file run, so if `test_file()` function was modified to check contents of the file , that would be applied to `:TestSuite` as well. If every test file needs to contain `Lab.script()`, then I think it would be ideal. You can use `readfile()` Vim function for that, which returns a List of lines.

sloria:
I've made the minor changes requested by @janko-m . Please let me know what approach you'd like to take re: detection of the correct test runner.

1. Use`test#javascript#runner` to explicitly declare the runner
2. Detect usage of lab by searching `Lab.script()` in the current file
3. Detect usage of lab by searching `package.json`
4. Some combination of the above

sloria:
Ah, just saw your last comment @janko-m . OK, I will look into that approach after work today.

sloria:
@janko-m I've made that change and removed `test#javascript#runner`, as it's no longer necessary. Please let me know if any further changes need to be made.

PR #186, opened 2017-04-26
add support for the Ginkgo framework of Go

blurrcat:
[Ginkgo](https://github.com/onsi/ginkgo) is a BDD-style Golang testing framework.

Although Ginkgo tests can be run by `go test`, `gotest.vim` is unable to build positions because of its BDD style. `ginkgo.vim` builds positions based on BDD keywords offered by Ginkgo: `It`, `Describe` and `Context`.

For your reference, Ginkgo cli offers two ways to run a subset of tests:
* regex on spec: `ginkgo --focus="should return 0"`
* regex on test file path: `ginkgo --regexScansFilePath=true --focus=package/some_test.go`

I'm fairly new to vim-script. Any suggestion will be appreciated, Thanks!

### DIFF
--- a/autoload/test/go/ginkgo.vim
+++ b/autoload/test/go/ginkgo.vim
@@ -11,7 +11,7 @@ function! test#go#ginkgo#test_file(file) abort
     if exists('g:test#go#runner')
       return g:test#go#runner == 'ginkgo'
     else
-      return executable('ginkgo')
+      return search("github.com/onsi/ginkgo") > 0
     endif
   endif
 endfunction
--- a/autoload/test/go/ginkgo.vim
+++ b/autoload/test/go/ginkgo.vim
@@ -7,13 +7,7 @@ let test#go#ginkgo#patterns = {
 \}
 
 function! test#go#ginkgo#test_file(file) abort
-  if fnamemodify(a:file, ':t') =~# g:test#go#ginkgo#file_pattern
-    if exists('g:test#go#runner')
-      return g:test#go#runner == 'ginkgo'
-    else
-      return search("github.com/onsi/ginkgo") > 0
-    endif
-  endif
+  return test#go#test_file('ginkgo', g:test#go#ginkgo#file_pattern, a:file)
 endfunction
 
 function! test#go#ginkgo#build_position(type, position) abort
--- a/autoload/test/go/ginkgo.vim
+++ b/autoload/test/go/ginkgo.vim
@@ -1,0 +1,50 @@
+if !exists('g:test#go#ginkgo#file_pattern')
+  let g:test#go#ginkgo#file_pattern = '\v[^_].*test\.go$'
+endif
+let test#go#ginkgo#patterns = {
+  \ 'test': ['\v^\s*It\("(.*)",', '\v^\s*Context\("(.*)",', '\v.*Describe\("(.*)",'],
+  \ 'namespace': [],
+\}
+
+function! test#go#ginkgo#test_file(file) abort
+  if fnamemodify(a:file, ':t') =~# g:test#go#ginkgo#file_pattern
+    if exists('g:test#go#runner')
+      return g:test#go#runner == 'ginkgo'
+    else
+      return executable('ginkgo')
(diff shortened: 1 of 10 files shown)


autoload/test/go/ginkgo.vim
```diff
@@ -0,0 +1,50 @@
...
+  let g:test#go#ginkgo#file_pattern = '\v[^_].*test\.go$'
+endif
+let test#go#ginkgo#patterns = {
+  \ 'test': ['\v^\s*It\("(.*)",', '\v^\s*Context\("(.*)",', '\v.*Describe\("(.*)",'],
+  \ 'namespace': [],
+\}
+
+function! test#go#ginkgo#test_file(file) abort
+  if fnamemodify(a:file, ':t') =~# g:test#go#ginkgo#file_pattern
+    if exists('g:test#go#runner')
```
codeinabox: Instead of doing this could you check the test file contains `github.com/onsi/ginkgo` which I am assuming has to be imported on all ginkgo tests? [This is what we've done for other runners](https://github.com/janko-m/vim-test/blob/master/autoload/test/php/phpspec.vim#L7)

blurrcat: `go test` and ginkgo test files share the same filename pattern. Even if we `search("github.com/onsi/ginkgo")` here, it is still necessary to use `test#go#runner` to force using one of the runners.

I think we should keep using `test#go#runner` and replace `executable("ginkto")` with `search("github.com/onsi/ginkgo")`. Is this OK?

codeinabox: The mantra for this plugin is "zero configuration required" so ideally users don't have to set a runner variable. What if ginkgo file matching search for "github.com/onsi/ginkgo" and the existing go runner file pattern only worked for files that don't contain it?

spec/go_runner_spec.vim
```diff
@@ -0,0 +1,38 @@
...
+  after
+    cd -
+  end
+
+  describe "when test#go#runner is not set"
+    after
+      call Teardown()
+    end
+    it "should use ginkgo if 'github.com/onsi/ginkgo' is found"
+      view ginkgo/normal_test.go
```
blurrcat: I wanted to have a unittest on `determine_runner` only, but I found that I have to vie w the file to make `search` work. Is there a better way?

autoload/test/go/ginkgo.vim
blurrcat: Now if the user does not set `go#runner`, it is determined by the presence of "github.com/onsi/ginkgo"

spec/ginkgo_spec.vim
```diff
@@ -0,0 +1,73 @@
+source spec/support/helpers.vim
+
+describe "Ginkgo"
+
+  before
+    let g:test#go#runner = 'ginkgo'
```
codeinabox: Could you  please add a test where is not set to test the search based runner detect

blurrcat: I couldn't find a way to parametrize spec in vspec.

I removed `let g:test#go#runner = 'ginkgo'` here so it tests the scenario when the user does not set `go#runner`, which is the default behavior. One more test is added to make sure TestNearest is not broken by the `search` call in `test_file`

spec/go_runner_spec.vim
codeinabox: It feels like the tests in this file should be against against a standard go test file making sure the auto-detection uses the go runner and not the ginkgo one. Otherwise looking good

blurrcat: Tests in this file do test against standard go test files. E.g.:
- [`view gotest/normal_test.go` on line 21](https://github.com/janko-m/vim-test/pull/186/files#diff-571e7dece54003e94224c6fda5045952R22) tests that gotest should be used if no ginkgo import is found
- [the for loop on line 30](https://github.com/janko-m/vim-test/pull/186/files#diff-571e7dece54003e94224c6fda5045952R30) tests that `go#runner` should be respected in both cases

Issue #187, opened 2017-05-05
Setting task runner doesn't work with javascript


lan-nguyen91:
this is what I have in my .vimrc

```
let test#strategy="basic"
let test#javascript#minitest#file_pattern=".spec\.js"
let test#javascript#runner="jasmine"
let g:test#runner_commands = ['jasmine']
```

I tried this too 
```
let test#javascript#runner="Jasmine"
let g:test#javascript#rspec#executable = 'Jasmine rspec'
```

The test ran, but it ran `mocha` as a runner, not sure what I did wrong.
I tried to capitalize letter `J` too, nothing works. I am using Neovim, and nvm, not sure if those two caught problems. 

The command that it prints is 
`:!node_modules/.bin/mocha test/controllers/client.spec.js` 

Can someone help,

Thanks.

janko-m:
@lan-nguyen91 Not all test runners implement `test#{language}#runner` detection, there is an open ticket for generalizing this so that it automatically works with any test runner: https://github.com/janko-m/vim-test/issues/185

janko-m:
@lan-nguyen91 As a temporary workaround you can try to set `test#javascript#mocha#file_pattern` to something that definitely won't match your test files, so that Mocha is skipped.

```vim
let test#javascript#mocha#file_pattern = 'mocha'
```

This should work unless there is some other JS runner after Mocha that will get detected by the same pattern.

We are also planning to modify all JavaScript test runners to only be activated if the corresponding Node package is declared in `package.json` (with the option to still manually set it via `test#javascript#runner` of course).

lan-nguyen91:
I remove the array of options in you vim file, restricting to jasmine only for now. But looking forward to your update mentioned above. This is very helpful, thanks a lot! 👍

codeinabox:
I could raise a PR to help with the detection. @janko-m are we thinking the package.json strategy?

janko-m:
@codeinabox Yes, I think that neither of the JavaScript test runners should be activated if they are not listed in `package.json`. In addition to that, we should also do https://github.com/janko-m/vim-test/issues/185, so that people can select a runner even if it's not listed in `package.json`. That would be global (in `autoload/test.vim`), and would need be careful not to break anything.

codeinabox:
What about yarn or as a first step should we just get it working for package.json?

Issue #185, opened 2017-04-24
Create a generic way of hardcoding the test runner


janko-m:
In some languages there are many test runners which use the same file pattern for test files (notably JavaScript and Python), and it's sometimes hard for test.vim to determine which test runner it should run. To work around that, for some of these test runners we are checking whether `test#{language}#runner` variables is declared and if it matches, to allow users to manually set which test runner they want to use.

In order to avoid the same `test#{language}#runner` logic in each of the test runners, and to allow users to always hardcode the test runner, we want to extract this logic out of the individual test runners and make it generic.

codeinabox:
I guess one way to get around this is check for runners in the project dev dependencies.

janko-m:
@codeinabox Yes, runner implementations should still attempt to do everything they can to determine the correct test runner, which includes checking dev dependencies. This would just be a way to override that and set the test runner manually, e.g. if the runner is installed only globally, or if test.vim has any other flaw in its logic.

codeinabox:
@janko-m based on your earlier points, on what level will a user set the runner - globally, per buffer or both?

> However, for the sake of consistency we can leave it as it is, and remove it once we implement #185. To clarify, after #185 I plan that we stop activating a JavaScript runner if only the pattern matches and the global executable exists (without it existing in package.json). Considering tha then users will be able to selecting runners manually (instead of having to rely on test.vim to always correctly detect it), we can require that pattern matches and the runner is present in package.json, or the runner is hardcoded via a variable.

janko-m:
@codeinabox Globally, I was thinking that the user can set the `g:test#runner` variable to the runner they want to use, in which case any pattern matching and other checks are skipped. We could then add a method which checks for both `g:test#runner` and the current `g:test#{language}#runner` variables, and in both cases returns the value (they would be treated the same).

codeinabox:
Sounds good. I could try picking this up in a couple of weeks when I am back from holidays.

codeinabox:
@janko-m I might be able to pick this up for JavaScript at the very least. Would this variable override the test file pattern or just the runner detection? I am thinking something like this 

```viml
function! test#javascript#mocha#test_file(file) abort
  return a:file =~# g:test#javascript#mocha#file_pattern
	  \ && (g:test#javascript#runner == 'mocha' || test#javascript#has_package('mocha') )
endfunction
```

janko-m:
[...]

Issue #49, opened 2015-07-31
PHPUnit and Behat configuration in subdirectories


codeinabox:
I've tested the PHPUnit and Behat runners and it works great except for when the configuration is in a subdirectory. How can I code `#build_args` to look for a phpunit.xml or behat.yml recursively?

janko-m:
You want to extract something out of `phpunit.xml` and `behat.yml`? I think the best way would be

```vim
glob('**/phpunit.xml')[0]
glob('**/behat.yml')[0]
```

These will throw an error if the glob is empty, so you may want to add an `if !empty(glob(...))`, or you can use `get(glob(...), 0)`, which will return "0" when the index is missing (that is, when the glob is empty).

codeinabox:
Don't want to extract anything, just want to properly set the configuration path if it's not in the root directory eg

> phpunit -c site/tests/phpunit.xml

I create a patch and try using hub to turn this issue into a pull request

janko-m:
Just a friendly reminder if you maybe have time to take a stab at this :)

codeinabox:
Still on my radar. One thing that I need to consider in my approach is to not look for phpuinit.xml in vendor directories

codeinabox:
Managed to make some headway on this but got two problems to solve. Firstly, how do I exclude things from the glob pattern. At the moment it returns all the below results but all I want is `site/test/phpunit.xml`

> site/test/phpunit.xml
site/vendor/aura/sql/tests/phpunit.xml
site/vendor/jalet/tvnu-config/phpunit.xml
site/vendor/mockery/mockery/examples/starship/phpunit.xml
site/vendor/mockery/mockery/tests/phpunit.xml

Secondly how do I make build args appear before the test file? ie `phpunit -c tests/phpunit.xml tests/NormalTest.php`

janko-m:
1. It doesn't seem that Vim's `glob()` has a negative pattern match. Maybe you could solve this by matching the normal glob and subtracting the `vendor/` glob (subtracting two arrays as sets, I don't know what's the function for that though).

2. You just need to define the `build_args()` function in the following way, that you prepend to the options array:
  ```rb
  function! test#php#phpunit#build_args(args) abort
    return ['-c test/phpunit.xml'] + a:args
  endfunction
  ```

janko-m:
Sorry for poking again, this is the only open issue :). Did you manage to get somewhere with this?

codeinabox:
[...]

PR #188, opened 2017-05-09
Detect the JavaScript runner based on the package.json

codeinabox:
This arises for #187 as based on this plugin's motto of zero configuration we should try to infer the runner from the package.json

### DIFF
--- a/autoload/test/javascript.vim
+++ b/autoload/test/javascript.vim
@@ -4,7 +4,7 @@ let test#javascript#patterns = {
 \}
 
 function! test#javascript#has_package(package) abort
-  exec 'silent grep! 'a:package .' package.json'
+  exec "silent grep! '\b".a:package."\b' package.json"
 
   return len(getqflist()) > 0
 endfunction
--- a/autoload/test/javascript.vim
+++ b/autoload/test/javascript.vim
@@ -2,3 +2,9 @@ let test#javascript#patterns = {
   \ 'test': ['\v^\s*%(it|test)\s*[( ]\s*%("|'')(.*)%("|'')'],
   \ 'namespace': ['\v^\s*%(describe|suite|context)\s*[( ]\s*%("|'')(.*)%("|'')'],
 \}
+
+function! test#javascript#has_package(package) abort
+  exec 'silent grep! 'a:package .' package.json'
+
+  return len(getqflist()) > 0
+endfunction
--- a/autoload/test/javascript.vim
+++ b/autoload/test/javascript.vim
@@ -4,7 +4,11 @@ let test#javascript#patterns = {
 \}
 
 function! test#javascript#has_package(package) abort
-  exec "silent grep! '\b".a:package."\b' package.json"
+  for line in readfile("package.json")
+    if line =~ '"'.a:package.'"'
+      return 1
+    endif
+  endfor
 
-  return len(getqflist()) > 0
+  return 0
 endfunction
--- a/autoload/test/javascript/jest.vim
+++ b/autoload/test/javascript/jest.vim
@@ -1,5 +1,5 @@
 if !exists('g:test#javascript#jest#file_pattern')
-  let g:test#javascript#jest#file_pattern = '(\v__tests__/.*.(js|jsx|coffee)|(\v.*(spec|test)\.(js|jsx|coffee))$'
+  let g:test#javascript#jest#file_pattern = '\v(__tests__/.*|(spec|test))\.(js|jsx|coffee)$'
 endif
 
 function! test#javascript#jest#test_file(file) abort
--- a/autoload/test/javascript/jest.vim
+++ b/autoload/test/javascript/jest.vim
@@ -1,5 +1,5 @@
 if !exists('g:test#javascript#jest#file_pattern')
-  let g:test#javascript#jest#file_pattern = '\v(__tests__/.*|\.(test|spec))\.(js|jsx|coffee)$'
+  let g:test#javascript#jest#file_pattern = '(\v__tests__/.*.(js|jsx|coffee)|(\v.*(spec|test)\.(js|jsx|coffee))$'
 endif
 
 function! test#javascript#jest#test_file(file) abort
--- a/autoload/test/javascript/jest.vim
+++ b/autoload/test/javascript/jest.vim
@@ -4,6 +4,7 @@ endif
 
 function! test#javascript#jest#test_file(file) abort
   return a:file =~# g:test#javascript#jest#file_pattern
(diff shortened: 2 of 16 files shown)


codeinabox:
@janko-m this is just a quick proof of concept. If you approve of this approach I'll update the existing JavaScript runners.

@lan-nguyen91 if you feel inclined could you try out this branch to see if the auto-detection for Jasmine works

codeinabox:
@janko-m this is working for all the JavaScript runners except intern. Any thoughts?

autoload/test/javascript.vim
```diff
@@ -2,3 +2,9 @@ let test#javascript#patterns = {
   \ 'test': ['\v^\s*%(it|test)\s*[( ]\s*%("|'')(.*)%("|'')'],
   \ 'namespace': ['\v^\s*%(describe|suite|context)\s*[( ]\s*%("|'')(.*)%("|'')'],
 \}
+
+function! test#javascript#has_package(package) abort
+  exec "silent grep! '\b".a:package."\b' package.json"
+
+  return len(getqflist()) > 0
```
janko-m: I think it's better to use `readfile("package.json")` here, and I would also search for the package name surrounded by double quotes, just to make sure we don't detect another package which includes the test runner name as a substring.

```vim
for line in readfile("package.json")
  if line =~ '".a:package."'
    return 1
  endif
endfor
return 0
```

autoload/test/javascript/mocha.vim
```diff
@@ -4,6 +4,7 @@ endif
 
 function! test#javascript#mocha#test_file(file) abort
   return a:file =~# g:test#javascript#mocha#file_pattern
+	&& test#javascript#has_package('mocha')
```
janko-m: The reason why `intern` tests aren't working is because you need to prepend this line with `\`, otherwise Vim won't consider this part of the previous line.

```vim
   return a:file =~# g:test#javascript#mocha#file_pattern
     \ && test#javascript#has_package('mocha')
```

Also, make sure you're using spaces and not tabs 😉

codeinabox:
Thanks for the feedback. As a fallback to the package.json detection would it make sense to also check if the runner is installed globally eg `which karma`?

autoload/test/javascript/mocha.vim
codeinabox: Have added an .editorconfig file to indicate vim files are 2 spaces

codeinabox:
@janko-m okay I've configured the behaviour to do the following:

```viml
    \ && (test#javascript#has_package('intern') || !empty(test#javascript#intern#executable()))
```
This will give precedence to what is in package.json otherwise fall back to what is returned by executable.

mjacobus:
@codeinabox 

pacakge.json
```javascript
{
  "name": "JestTest",
  "version": "1.0.0",
  "main": "index.js",
  "license": "MIT",
  "dependencies": {
    "jest": "^20.0.4"
  },
  "scripts": {
      "test": "jest"
  }
}
```

spec/sum.test.js still runs with mocha.

codeinabox:
@mjacobus the current behaviour of the jest test runner is only to recognise tests in the directory `__tests__`. Reading [through the Jest documentation](https://facebook.github.io/jest/docs/configuration.html#testmatch-array-string) we also need to allow for files names like you describe, I will add this to my PR.

mjacobus:
Maybe it could depend on the file suffix rather than the directory it is in.

codeinabox:
I'll update this PR to support Jest's default pattern of `[ '**/__tests__/**/*.js?(x)', '**/?(*.)(spec|test).js?(x)' ]`

codeinabox:
@mjacobus I have update the regex for Jest. Let me know if this works for you.

autoload/test/javascript/jest.vim
```diff
@@ -1,9 +1,10 @@
 if !exists('g:test#javascript#jest#file_pattern')
-  let g:test#javascript#jest#file_pattern = '\v(__tests__/.*|\.(test|spec))\.(js|jsx|coffee)$'
+  let g:test#javascript#jest#file_pattern = '(\v__tests__/.*.(js|jsx|coffee)|(\v.*(spec|test)\.(js|jsx|coffee))$'
```
janko-m: I think this way `(` characters are taken literally, because `\v` is applied only after, the following regex is in my opinion better and shorter:

```
\v(__tests__/.*|(spec|test))\.(js|jsx|coffee)
```

codeinabox: Thanks for the quick feedback. Will update.

autoload/test/javascript/mocha.vim
```diff
@@ -4,6 +4,7 @@ endif
 
 function! test#javascript#mocha#test_file(file) abort
   return a:file =~# g:test#javascript#mocha#file_pattern
+	  \ && (test#javascript#has_package('mocha') || !empty(test#javascript#mocha#executable()))
```
janko-m: In some JavaScript test runners like `mocha`, `test#javascript#{runner}#executable()` will fallback to a global executable if it doesn't find the one in `node_modules/.bin`, so the `!empty(test#javascript#mocha#executable())` conditional will always evaluate to true.

However, for the sake of consistency we can leave it as it is, and remove it once we implement https://github.com/janko-m/vim-test/issues/185. To clarify, after https://github.com/janko-m/vim-test/issues/185 I plan that we stop activating a JavaScript runner if only the pattern matches and the global executable exists (without it existing in `package.json`). Considering tha then users will be able to selecting runners manually (instead of having to rely on test.vim to always correctly detect it), we can require that pattern matches and the runner is present in `package.json`, or the runner is hardcoded via a variable.

Just wanted to have a stream of thought, nothing to change here now.

autoload/test/javascript.vim
```diff
@@ -2,3 +2,13 @@ let test#javascript#patterns = {
   \ 'test': ['\v^\s*%(it|test)\s*[( ]\s*%("|'')(.*)%("|'')'],
   \ 'namespace': ['\v^\s*%(describe|suite|context)\s*[( ]\s*%("|'')(.*)%("|'')'],
 \}
+
+function! test#javascript#has_package(package) abort
+  for line in readfile("package.json")
```
codeinabox: We need to be careful when searching for package.json recursively as we need to make the vendor directory ie `node_modules` is excluded.

I am inclined to leave this out of this PR as if we add recursive detection of package.json then we need to add recursive detection of test runners. Subdirectory support feels like a bigger feature, I know we parked it for PHP runners #49

janko-m: Good point, let's delay it for a future discussion then.

PR #190, opened 2017-05-21
Use regex similar to jest official doc

sharils:
This original regex is mentioned in the testRegex section package.json example in https://facebook.github.io/jest/docs/en/getting-started.html#using-typescript

---
This change is [Reviewable](https://reviewable.io/reviews/janko-m/vim-test/190)


janko-m:
Looks good, except that the leading `/` should be excluded, as test.vim by default works with relative paths.

sharils:
Sorry, I broke the test, I'm gonna fix the tests and send another PR. :(

janko-m:
Thanks!

PR #191, opened 2017-05-23
Add Racket's rackunit to vim-test

joereynolds:
This adds support for Racket's built-in unit testing module, Rackunit.

Unfortunately a lot of the options available (running suites etc...) aren't available through the CLI meaning we can only really do :TestFile (hence the very slim PR)

Let me know if there are any problems.

Thanks! :D
Joe


autoload/test/racket/rackunit.vim
```diff
@@ -0,0 +1,23 @@
...
+  return a:args
+endfunction
+
+function! test#racket#rackunit#executable() abort
+  return 'racket'
+endfunction
+
+function! s:nearest_test(position)
+  return 'This functionality is not available in rackunit'
+endfunction
```
janko-m: This function doesn't need to be defined, in other runners it's defined as a custom helper function for that specific runner, but here it's not used.

plugin/test.vim
```diff
@@ -27,6 +27,7 @@ call s:extend(g:test#runners, {
   \ 'Lua':        ['Busted'],
   \ 'PHP':        ['Codeception', 'PHPUnit', 'Behat', 'PHPSpec', 'Kahlan', 'Peridot'],
   \ 'Perl':       ['Prove'],
+  \ 'Racket':     ['Rackunit'],
```
janko-m: I would prefer if it was named `RackUnit`, that's what the command name would be if the user chooses to generate it.

README.md
```diff
@@ -20,6 +20,7 @@ Currently the following testing frameworks are supported:
 | **PHP**        | Behat, Codeception, Kahlan, Peridot, PHPUnit, PHPSpec | `behat`, `codeception`, `kahlan`, `peridot`, `phpunit`, `phpspec` |
 | **Perl**       | Prove                                                 | `prove`                                                           |
 | **Python**     | Django, Nose, Nose2, PyTest, PyUnit                   | `djangotest`, `djangonose` `nose`, `nose2`, `pytest`, `pyunit`    |
+| **Racket**     | Rackunit                                              | `rackunit`                                                        |
```
janko-m: Here you can also call it "RackUnit", since that's the official name.

PR #193, opened 2017-05-23
FIX better support for options in maventest

### LINKED ISSUE #194: maventest.vim cannot support options
java/maventest.vim cannot currently support options that users may want to pass to the test runner. The arguments are processed by [`test#java#maventest#build_args`](https://github.com/janko-m/vim-test/blob/master/autoload/test/java/maventest.vim#L26). 

The function accepts a list of arguments which contains both the options that the user may set in `test#java#maven#options` and the test file name. The next line simply prepends a cli parameter for the test runner to the list of arguments without any checks.

For example if a user wants to use a custom `pom.xml` file for maven configuration they must set
`let test#java#maventest#options = '-f /path/to/pom.xml'`

Once a test is executed, the resulting command becomes (which is not valid):
`$ mvn test -Dtest=-f /path/to/pom.xml TestName`

The correct command should be:
`$ mvn test -f /path/to/pom.xml -DTest=TestName`

mvalkon:
maventest.vim could not support the vim-test way of adding options for a test
runner. For example if a user tried to pass `-f /path/to/pom.xml` as an option
parameter as follows:

`let test#java#maventest#options = '-f /path/to/pom.xml'`

the script would end up calling

`mvn test -Dtest=-f /path/to/pom.xml TestFileName`

which cannot work. Corrected the argument handling to support any number of
option parameters. The assumption is that 0 to 1 test names are passed to the
plugin. The example above now results to:

`mvn test -f /path/to/pom.xml -Dtest=TestFileName`

Fixes #194

### DIFF
--- a/autoload/test/java/maventest.vim
+++ b/autoload/test/java/maventest.vim
@@ -24,7 +24,7 @@ function! test#java#maventest#build_position(type, position) abort
 endfunction
 
 function! test#java#maventest#build_args(args) abort
-    return a:args
+  return a:args
 endfunction
 
 function! test#java#maventest#executable() abort
--- a/autoload/test/java/maventest.vim
+++ b/autoload/test/java/maventest.vim
@@ -6,26 +6,26 @@ function! test#java#maventest#test_file(file) abort
   return a:file =~? g:test#java#maventest#file_pattern
 endfunction
 
+
 function! test#java#maventest#build_position(type, position) abort
   let filename = fnamemodify(a:position['file'], ':t:r')
 
   if a:type == 'nearest'
     let name = s:nearest_test(a:position)
     if !empty(name)
-      return [name]
+      return ['-Dtest=' . name]
     else
-      return [filename]
+      return ['-Dtest=' . filename]
     endif
   elseif a:type == 'file'
-    return [filename]
+    return ['-Dtest=' . filename]
   else
     return []
   endif
 endfunction
 
 function! test#java#maventest#build_args(args) abort
-  let args = ['-Dtest='] + a:args
-  return [join(args, "")]
+    return a:args
 endfunction
 
 function! test#java#maventest#executable() abort
--- a/autoload/test/java/maventest.vim
+++ b/autoload/test/java/maventest.vim
@@ -6,7 +6,6 @@ function! test#java#maventest#test_file(file) abort
   return a:file =~? g:test#java#maventest#file_pattern
 endfunction
 
-
 function! test#java#maventest#build_position(type, position) abort
   let filename = fnamemodify(a:position['file'], ':t:r')
 
--- a/spec/maventest_spec.vim
+++ b/spec/maventest_spec.vim
@@ -7,7 +7,6 @@ describe "Maven"
   end
 
   after
-    unlet! g:test#java#maventest#options
     call Teardown()
     cd -
   end
@@ -21,8 +20,7 @@ describe "Maven"
 
   it "runs file tests with user provided options"
     view MathTest.java
-    let g:test#java#maventest#options = '-f pom.xml'
-    TestFile
+    TestFile -f pom.xml
 
     Expect g:test#last_command == 'mvn test -f pom.xml -Dtest=MathTest'
   end
@@ -43,8 +41,7 @@ describe "Maven"
 
   it "runs a test suite with user provided options"
     view MathTest.java
-    let g:test#java#maventest#options = '-X -f pom.xml -DcustomProperty=5'
-    TestSuite
+    TestSuite -X -f pom.xml -DcustomProperty=5
 
     Expect g:test#last_command == 'mvn test -X -f pom.xml -DcustomProperty=5'
   end
--- a/spec/maventest_spec.vim
+++ b/spec/maventest_spec.vim
@@ -7,6 +7,7 @@ describe "Maven"
   end
 
   after
+    unlet! g:test#java#maventest#options
     call Teardown()
     cd -
   end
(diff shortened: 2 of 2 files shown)


mvalkon:
@codeinabox will do

codeinabox:
I've added a spec for Maven so please rebase against master and add your tests on top of that.

mvalkon:
@codeinabox That is very kind of you, thanks a lot for creating it. I think it's good enough for me to write a spec for the functionality in this pr. Much appreciated 👍

mvalkon:
@codeinabox thanks for the spec, made everything a lot easier! Could you take a look at the changes?

spec/support/helpers.vim
```diff
@@ -20,5 +20,5 @@ endfunction
 
 function! Teardown() abort
   bufdo! bdelete!
-  unlet! g:test#last_command g:test#last_position g:test#last_strategy
+  unlet! g:test#last_command g:test#last_position g:test#last_strategy g:test#java#maventest#options
```
codeinabox: Could you move this into the `after` block for `maventest_spec.vim`

mvalkon: Yup makes sense, done.

autoload/test/java/maventest.vim
```diff
@@ -24,8 +24,20 @@ function! test#java#maventest#build_position(type, position) abort
 endfunction
 
 function! test#java#maventest#build_args(args) abort
-  let args = ['-Dtest='] + a:args
-  return [join(args, "")]
+  " In order to support custom options passed to mvn we need split
+  " the options from the test name (if one is passed).
+  let test_name = ''
+  let opts = []
+  let user_options = get(g:, 'test#java#maventest#options', '')
```
janko-m: You shouldn't need to read the `test#java#maventest#options` variable, its value is already merged into `a:args`.

mvalkon: Yes and that's inherently _the issue_ here.

For maven we need to insert a cli parameter `-DTest=` before the test name but after any custom options the user gives, so I'm comparing which items in `a:args` are also passed as options, and which (if any) is the test name. 

I initially had something like this (pardon my poor vimscript knowledge):
```
let args = a:args[:-2] + ['-Dtest=' . a:args[-1]]
```
The above takes all items in `a:args` until the last one (which I assume is the test file) and appends a concatenation of `-Dtest=TestName`. If we use the above and want to pass `-f /path/to/pom.xml` for mvn it will work with `:TestFile` and  `:TestNearest`. It would break `:TestSuite` though as it would eventually result in
```
$ mvn -f /path/to/pom.xml -Dtest=/path/to/pom.xml
```

I haven't really worked with vimscript before so there might be a better way to achieve this. Suggestions are welcome.

codeinabox: Would a simpler solution be just this?

```viml
if !exists('g:test#java#maventest#file_pattern')
  let g:test#java#maventest#file_pattern = '\v^.*[Tt]est\.java$'
endif

function! test#java#maventest#test_file(file) abort
  return a:file =~? g:test#java#maventest#file_pattern
endfunction

function! test#java#maventest#build_position(type, position) abort
  let filename = fnamemodify(a:position['file'], ':t:r')

  if a:type == 'nearest'
    let name = s:nearest_test(a:position)
    if !empty(name)
      return ['-Dtest=' . name]
    else
      return ['-Dtest=' . filename]
    endif
  elseif a:type == 'file'
    return ['-Dtest=' . filename]
  else
    return ['-Dtest=']
  endif
endfunction

function! test#java#maventest#build_args(args) abort
  return a:args
endfunction

function! test#java#maventest#executable() abort
  return 'mvn test'
endfunction

function! s:nearest_test(position) abort
  let name = test#base#nearest_test(a:position, g:test#java#patterns)
  return escape(join(name['namespace'] + name['test'], '#'), "#")
endfunction

```

mvalkon: @codeinabox that certainly looks cleaner. Would you like to commit to this PR?

codeinabox: You will need to modify your PR as I can't.

codeinabox: Also what should the TestSuite behaviour be? Could you also write a spec for that please

mvalkon: @codeinabox done, it's quite a bit nicer now, thanks! The `TestSuite` behavior should be clear from the spec.

autoload/test/java/maventest.vim
```diff
@@ -6,26 +6,26 @@ function! test#java#maventest#test_file(file) abort
   return a:file =~? g:test#java#maventest#file_pattern
 endfunction
 
+
```
janko-m: Extra newline that probably wasn't intentional 😛

spec/maventest_spec.vim
```diff
@@ -18,11 +19,34 @@ describe "Maven"
     Expect g:test#last_command == 'mvn test -Dtest=MathTest'
   end
 
+  it "runs file tests with user provided options"
+    view MathTest.java
+    let g:test#java#maventest#options = '-f pom.xml'
+    TestFile
```
janko-m: Instead of assigning the `g:test#java#maventest#options` variable, you can pass these options directly to the command:

```vim
view MathTest.java
TestFile -f pom.xml
```

All test.vim commands accept command line options that they just forward to the test runner (they get merged into to `a:args` in `test#{language}#{runner}#build_args`), in the same way that that options from the `g:test#{language}#{runner}#options` variable get merged. In other words, these two ways of passing arguments are equivalent.

spec/maventest_spec.vim
```diff
@@ -7,6 +7,7 @@ describe "Maven"
   end
 
   after
+    unlet! g:test#java#maventest#options
```
janko-m: Without the variable this call wouldn't be needed anymore.

spec/maventest_spec.vim
mvalkon: Nice to know, thanks for the suggestion. Fixed according to comments.

mvalkon:
I squashed some of the intermediary commits to keep things a little cleaner. Let me know if you want me to squash the last update as well.

autoload/test/java/maventest.vim
```diff
@@ -12,20 +12,19 @@ function! test#java#maventest#build_position(type, position) abort
...
+    return ['-Dtest=' . filename]
   else
     return []
   endif
 endfunction
 
 function! test#java#maventest#build_args(args) abort
-  let args = ['-Dtest='] + a:args
-  return [join(args, "")]
+    return a:args
```
codeinabox: Indent two spaces please

mvalkon: Sorry, my shiftwidth is globally set to 4 :) should be fixed now

codeinabox:
@mvalkon thank you for your work! Feels a big improvement now that we have specs for how this should behave.

mvalkon:
@codeinabox you can thank yourself for writing the initial spec and helping with the implementation. All in all I'd like to thank you and @janko-m for quick response time. I think quick and valuable feedback is the most crucial thing when someone contributes to a specific open source project for the first time.

PR #195, opened 2017-05-24
Add spec for Maven test runner

codeinabox:


codeinabox:
@janko-m for some strange the nearest spec fails on Travis but passes fine on my local machine.

@mvalkon let me know if the example test and spec if a sufficient base for your PR #193

Issue #14, opened 2015-01-29
Feature Request: Failure Navigation


alex-netquity:
It would be convenient if there was an option to feed failures to the quickfix window and to be able to navigate to the failure lines the usual quickfix mappings. I've seen other vim testing plugins do this.

janko-m:
There already are even two ways to do it in vim-test:

1. Instruct vim-test to use the "make" strategy for running tests. That will select the appropriate Vim compiler (if it's available), and run the command with Vim's built-in `:make`, which loads the errors into a quickfix.

   ```vim
  let test#strategy = 'make'
   ```

2. Instruct vim-test to use the "dispatch" strategy for running tests. This uses [Dispatch.vim](https://github.com/tpope/vim-dispatch)'s `:Make` command, which is like an asynchronous `:make`.

   ```vim
  let test#strategy = 'dispatch'
   ```

It's included in the readme, but I will try to emphasize it more and explain what each strategy does.

alex-netquity:
From what I can tell, using either `make` or `dispatch` feeds the entire result set into the quickfix window. Once there, it's possible to traverse the output as one might expect. This is different from what I meant.

I've seen other testing plugins feed the results into the quickfix window and allow the user to jump to the specific line on which each test failed using `:cnext` and `:cprevious`. These commands would navigate to the lines in the code where the failures occurred rather than navigating through the test output.

Is similar functionality available through vim-test?

janko-m:
What test runner are you using with vim-test? When a quickfix list is loaded with the output, but the failures are not "linked" to the file locations, that happens when the "compiler" for that test runner doesn't exist. What Vim compilers do is they are able to read the test output, and create a quickfix list out of it (which can be traverse with `:cnext` and `:cprevious`).

Unfortunately, if the compiler doesn't exist for your test runner, vim-test can't provide it for you, because I would like to limit the scope of vim-test.

janko-m:
Also, which other testing plugins do the quickfix? I would just like to take a look.

alex-netquity:
I am using py.test for my test runner.

I can't recall exactly which testing plugin it was. It was some time ago. I believe it was https://github.com/reinh/vim-makegreen with the following command: `map <leader>dt :set makeprg=python\ manage.py\ test\|:call MakeGreen()<CR>`. At the time I was just using the Django test runner and not py.test.

janko-m:
For Python there is a "pyunit" Vim compiler, maybe it works for py.test. Try

```vim
[...]

Issue #166, opened 2017-01-08
Too early compiler choice


przepompownia:
With the below configuration 
```
" vim-test
let test#strategy = 'dispatch'

" vim-dispatch
let g:dispatch_compilers = {}
let g:dispatch_compilers.phpunit = 'phpunit2'
```
when I run, for example, `:TestNearest`, then `:Dispatch` gets `./vendor/bin/phpunit` into the first argument so `phpunit` compiler is ignored.

Although I found a workaround by set this argument manually
```
let test#php#phpunit#executable = 'phpunit'
```
I think that `test#php#phpunit#executable()` first should get the proper value from the compiler definition if that exist and if it make sense. It seems that searching the executable should be the responsibility of a particular _compiler_. 

Maybe this issue is more generic - not specific for `phpunit` and independent of `vim-dispatch`.

janko-m:
The thing is that test.vim isn't generally related to Vim compilers. You can use a strategy that uses compilers under the hood (dispatch.vim), or a strategy which doesn't use them at all.

The reason why test.vim cannot utilize compilers is because: (a) it would have to re-implement searching trough compiler (because it cannot rely on dispatch.vim being installed), and (b) even if it did, using the command declared by the compiler isn't necessarily what users want to be run.

For example, I'm assuming that running `phpunit` isn't equivalent to running `./vendor/bin/phpunit`. If test.vim would choose the compiler command, users wouldn't be able to choose to run `./vendor/bin/phpunit` and still use the `phpunit2` compiler.

However, you can still configure dispatch.vim to use the `phpunit2` compiler without changing anything in test.vim. The `g:dispatch_compilers` dictionary has a double functionality: in addition to the dictionary value being a compiler name, it can also be a blank string, and then dispatch.vim will treat the key as a prefix that should be subtracted, and continue the compiler lookup without that prefix.

In your case you want to ignore `./vendor/bin/` from the command, so you can specify:

```vim
let g:dispatch_compilers = {}
let g:dispatch_compilers['./vendor/bin/'] = ''
let g:dispatch_compilers['phpunit'] = 'phpunit2'
```

Then `./vendor/bin/phpunit ...` command should be transformed into `phpunit ...`, which should then activate the `phpunit2` compiler.

PR #199, opened 2017-06-26
Update docs to include information about the strategies that use quickfix

codeinabox:
Arises from #14

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,6 +131,16 @@ Even if you're using a strategy other than Dispatch.vim, it's recommended you ha
 Dispatch.vim installed, as the strategies will automatically use to determine the 
 correct compiler, ensuring the test output is correctly parsed for the quickfix window.
 
+If the test command prefix doesn't match compiler's `makeprg` then use the 
+`g:dispatch_compilers` variable. For example if your test command was `./vendor/bin/phpunit` 
+but you wanted to use the phpunit2 compiler you could do.
+
+```vim
+let g:dispatch_compilers = {}
+let g:dispatch_compilers['./vendor/bin/'] = ''
+let g:dispatch_compilers['phpunit'] = 'phpunit2'
+```
+
 ### Custom Strategies
 
 Strategy is a function which takes one argument – the shell command for the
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -117,6 +117,18 @@ disable this:
 let g:test#preserve_screen = 1
 ```
 
+### Quickfix Strategies
```
janko-m: Empty line missing between the title and the first paragraph.

README.md
```diff
@@ -117,6 +117,18 @@ disable this:
...
+If you want your test results to appear in the quickfix window, use one of the 
+following strategies:
+
+ * Make
+ * Neomake
+ * MakeGreen
+ * Dispatch.vim
+
+When Dispatch.vim is installed, it will be used to determine the compiler and 
+errorformat settings for the test runner.
```
janko-m: I would prefer for this paragraph to be something like:

> Even if you're using a quickfix strategy other than Dispatch.vim, it's still recommended you have Dispatch.vim installed, as the strategies will automatically use it to find and apply the correct Vim compiler, ensuring the test output is correctly parsed for the quickfix window.

There is no need to mention `errorformat` when we already mention "compiler", as setting `errorformat` is the only thing the compiler does (other than setting `makeprg` which we don't use).

codeinabox:
Have updated based on feedback. Do we need to explain to users that not all compilers are supported out of the box by Dispatch?

codeinabox:
@janko-m any further feedback on this?

janko-m:
It would be great to explain what you mentioned, and that compilers come from the Vim distribution and plugins, not from Dispatch.vim.

It would also be good to document the g:dispatch_compilers variable, as it's really useful when the test command prefix doesn't match compiler's makeprg. But if you're not sure what that variable does exactly (it took myself some time to understand), I can fill in that part.

codeinabox:
Using #166 as reference I have added an example of g:dispatch_compilers. Feel free to tweak anything I've written.

codeinabox:
@janko-m have ready through it I think is thorough enough but let me know if I have missed anything and feel free to alter

codeinabox:
@janko-m are you happy for me to merge this?

PR #201, opened 2017-07-19
Fix readme formatting.

NJichev:
Go code example for specific runner was missing
3 ticks which bugged the whole ruby paragraph.

Fixed by adding back the 3 ticks to the vim example.


janko-m:
Thanks!

PR #202, opened 2017-07-21
Let test be in front or at the end.

lyz-code:
As stated [here](https://pytest.readthedocs.io/en/reorganize-docs/new-docs/user/naming_conventions.html).

Taken as reference [the nose modification](https://github.com/janko-m/vim-test/pull/32/files)

### DIFF
--- a/autoload/test/python/pytest.vim
+++ b/autoload/test/python/pytest.vim
@@ -1,5 +1,5 @@
 if !exists('g:test#python#pytest#file_pattern')
-  let g:test#python#pytest#file_pattern = '^test.*\.py$'
+  let g:test#python#pytest#file_pattern = '\v(^|[\b_\.-])test.*\.py$'
 endif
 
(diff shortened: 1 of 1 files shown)


codeinabox:
Thank you for the PR! Could you please update the spec and fixture located in the `spec` folder

lyz-code:
Could you please tell me what to update in the spec and fixture?

I don't see anything that needs changing

codeinabox:
Sorry I meant to say could you please check the directory and file name structure for the nose fixtures are representative of the naming convention if they aren't already. Aside from that LGTM 👍

lyz-code:
If I understood correctly what you are saying, yes, pytest naming convention is compatible with the nose one. As stated on the link of the description, what Pytest says is that the filenames should be `test_filename.py` or `filename_test.py`.

I don't know if that answers your statement :S

PR #204, opened 2017-07-29
Add initial support for Elm and elm-test

bobwhitelock:
This PR adds initial basic support for [Elm](http://elm-lang.org/) and [elm-test](https://github.com/elm-community/elm-test) to vim-test. Thanks :slightly_smiling_face:

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,6 +12,7 @@ Currently the following testing frameworks are supported:
 | **Clojure**    | Fireplace.vim                                         | `fireplacetest`                                                   |
 | **Crystal**    | Crystal                                               | `crystalspec`                                                     |
 | **Elixir**     | ESpec, ExUnit                                         | `espec`, `exunit`                                                 |
+| **Elm**        | elm-test                                              | `elmtest`                                                         |
 | **Erlang**     | CommonTest                                            | `commontest`                                                      |
 | **Go**         | Ginkgo, Go                                            | `ginkgo`, `gotest`                                                |
 | **Java**       | Maven                                                 | `maventest`                                                       |
--- a/autoload/test/elm/elmtest.vim
+++ b/autoload/test/elm/elmtest.vim
@@ -1,0 +1,30 @@
+
+if !exists('g:test#elm#elmtest#file_pattern')
+  let g:test#elm#elmtest#file_pattern = '\vtests?\/.*\.elm$'
+endif
+
+function! test#elm#elmtest#test_file(file) abort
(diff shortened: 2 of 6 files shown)


bobwhitelock:
Thanks for the comments.

Regarding `doc/test.txt`, where do the runner commands get defined? The logical name in this case would be an `ElmTest` command, however this will conflict with [a command from the elm-vim plugin](https://github.com/ElmCast/elm-vim/#usage), which presumably most people who would use this runner will also be using. So I'm not sure what should be done in this case.

I'll look in to adding tests soon, I've not tested VimScript before.

codeinabox:
Runner commands are not enabled by default, you need to opt-in ie

```viml
let g:test#runner_commands = ['ElmTest']
```

Regarding specs and fixtures, if you use the existing ones as reference it should be straightforward but let me know if you get stuck.

bobwhitelock:
OK I've had another look and added those things, let me know what you think

spec/fixtures/elmtest/elm-package.json
```diff
@@ -0,0 +1,14 @@
...
+    "version": "4.0.0",
+    "summary": "Example Project",
+    "repository": "https://github.com/user/project.git",
+    "license": "BSD-3-Clause",
+    "source-directories": [
+        "src"
+    ],
+    "exposed-modules": [],
+    "dependencies": {
+        "elm-lang/core": "5.0.0 <= v < 6.0.0"
```
codeinabox: This file doesn't seem to be needed in the fixtures so you can remove

spec/fixtures/elmtest/tests/elm-package.json
```diff
@@ -0,0 +1,16 @@
...
+    "repository": "https://github.com/elm-community/elm-test.git",
+    "license": "BSD-3-Clause",
+    "source-directories": [
+        "."
+    ],
+    "exposed-modules": [],
+    "native-modules": true,
+    "dependencies": {
+        "elm-lang/core": "5.0.0 <= v < 6.0.0",
+        "elm-community/elm-test": "4.0.0 <= v < 5.0.0"
```
codeinabox: Like wise you can remove this file.

bobwhitelock:
OK, removed those files - the test command won't actually work without them, but you're right that they're not needed for the tests to pass.

codeinabox:
@janko-m any further feedback or are you happy for me to merge?

janko-m:
Looks great, thank you!

PR #205, opened 2017-07-30
Add support for swift test

gfontenot:
Swift Package Manager ships a test runner that is accessible via `swift test`.
This commit adds support for working with `swift test` via vim-test.

There's definitely some fragility in here still that might need to get worked
out over time. Off the top of my head, some potential issues:

- Users might locate their tests in a directory named something other than
  "Tests". Doing so would break the test file detection (and the module name
  parsing).
- If users use a test lib like Quick, this won't properly parse module info
  (and filtering Quick tests with swift test doesn't work at all).

That being said, I think this is a good start, and getting this into the
plugin would mean being able to deploy fixes for these issues by tweaking the
regexes instead of needing to write this whole thing from scratch.

### DIFF
--- a/autoload/test/swift/swiftpm.vim
+++ b/autoload/test/swift/swiftpm.vim
@@ -1,0 +1,67 @@
+if !exists('g:test#swift#swiftpm#file_pattern')
+  let g:test#swift#swiftpm#file_pattern = '\v^Tests\/.*\.swift$'
+endif
+
+function! test#swift#swiftpm#test_file(file) abort
+  return a:file =~# g:test#swift#swiftpm#file_pattern
+endfunction
+
+function! test#swift#swiftpm#build_position(type, position) abort
+  if a:type ==# 'nearest'
+    let l:module = s:parse_module_info(a:position)
+    let l:testcase = s:parse_case_info(a:position)
+    let l:nearest = s:parse_nearest_test_info(a:position)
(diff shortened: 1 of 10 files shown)


gfontenot:
I'm not sure why CI is failing here... I can run these tests locally and see them pass, and I can also run the plugin against actual code and see it perform as expected 😕

autoload/test/swift/swiftpm.vim
```diff
@@ -0,0 +1,67 @@
...
+      let l:result = l:match
+      break
+    endif
+  endfor
+
+  return l:result
+endfunction
+
+function! s:parse_nearest_test_info(position)
+  let l:info = g:test#base#nearest_test(a:position, g:test#swift#patterns)
```
codeinabox: I think you need to put this instead judging by the error

```viml
let l:info = test#base#nearest_test(a:position, g:test#swift#patterns)
```

gfontenot:
Fixed CI and added Swift support to the README. I didn't see anywhere obvious to update `doc/test.txt`. Happy to do that if you point me to what should be updated.

codeinabox:
For doc/test.txt it's just a case of adding something like this amongst all the runner commands

```
                                                *test-: SwiftPM*
:SwiftPM [args]               Uses the `swift test` command.
```

gfontenot:
Ah! My bad. Added that there now. I don't know why I didn't see that section.

janko-m:
Looks great, thank you very much!

Issue #206, opened 2017-08-01
Support maventest files staring with Test


shane-axiom:
Test*.java is a valid Maven Surefile test filename (the first option listed here):

http://maven.apache.org/surefire/maven-surefire-plugin/examples/inclusion-exclusion.html

However the `maventest` filepatten is only allowing files which end in Test.java:

https://github.com/janko-m/vim-test/blob/master/autoload/test/java/maventest.vim#L2

Can both variants be allowed? Probably the other two (`*Tests.java` and `*TestCase.java`) should be allowed as well?

codeinabox:
Sure, supporting both should be an easy change if you would like to raise a PR. It should just be case of creating a new test in maventest_spec.vim, for example

```
  it "runs file test with wildcard"
    view MathTestCase.java
    TestFile

    Expect g:test#last_command == 'mvn test -Dtest=MathTestCase'
  end
```

And then updating the regular expression in `g:test#java#maventest#file_pattern`. Let me know if you need any help.

PR #208, opened 2017-08-01
Make maventest match all surefire filename patterns (janko-m/vim-test#206)

shane-axiom:
maventest should match all Maven Surefire test filename patterns.

```
Test*.java
*Test.java
*Tests.java
*TestCase.java
```

http://maven.apache.org/surefire/maven-surefire-plugin/examples/inclusion-exclusion.html

Fixes https://github.com/janko-m/vim-test/issues/206.


codeinabox:
Thank you for the patch, very much appreciated!

PR #210, opened 2017-08-18
add es6 backticks support

nilsboy:


codeinabox:
Thank you for the PR! Is it possible to add an example of a file with back ticks to one of the existing JavaScript test fixtures so we can have an automated test for this.

janko-m:
@nilsboy Thanks for the patch! I added a test for the backticks while merging.

PR #211, opened 2017-08-19
Improve mocha suite test running

rumblesan:
Pulled out separately from the MochaWebpack branch I'm working on. I believe this better handles getting mocha to find all the files in a suite

recurses into sub dirs
tries both test and tests


rumblesan:
so I found that the tests wouldn't be picked up unless I passed in a folder. And as the `mocha#file_pattern` supports `test/` and `tests/` folders then it made sense for me to check for both of them.

Issue #213, opened 2017-08-21
/bin/bash: jest: command not found


Akhetopnu:
None of the vim-test commands work *unless* the test files are located in the `test` folder. If I want to execute tests in any random subdirectory in the project, vim-test says

:TestFile
```
jest src/item.test.js
/bin/bash: jest: command not found

shell returned 127

Press ENTER or type command to continue
```
I have no idea where does it get this `jest` from, the only dependency I have installed is `mocha`. I don't have `jest` installed anywhere on my computer.
The README file says `vim-test` is supposed to 'automagically detect' javascript test framework by reading `package.json`.

When I run `npm test` mocha finds the `item.test.js` file and runs the test.

Project tree (node_modules exists, simply not included here in the tree)
```
├── package.json
└── src
    └── item.test.js

1 directory, 2 files
```

package.json
```{
  "name": "vimtest",
  "version": "1.0.0",
  "description": "",
  "main": "index.js",
  "scripts": {
    "test": "mocha --recursive src/**/*.test.js"
  },
  "keywords": [],
  "author": "",
  "license": "ISC",
  "dependencies": {
    "mocha": "^3.5.0"
  }
}
```

item.test.js
```
const assert = require('assert');

describe('item', () => {
   it('should pass', () => {
      assert.strictEqual(true, false, 'true !== false');
   });
});
```

Is there something blatantly obvious here that I'm missing?

codeinabox:
Thank you for the detailed report. Let me try recreating the bug on my machine and I'll get back to you.

janko-m:
@codeinabox I think the following logic that all JavaScript runners is not entirely correct: https://github.com/janko-m/vim-test/blob/374218018c133077b899246878879de7e49a2427/autoload/test/javascript/jest.vim#L5-L8

Specifically, the `!empty(test#javascript#{runner}#executable())` condition isn't correct, because it will always evaluate to true,

codeinabox:
I am happy to raise a PR for this as I recently had a problem with runner detection.

janko-m:
@codeinabox @Akhetopnu So, https://github.com/janko-m/vim-test/pull/214 was merged, which means that now, instead of having a random JS runner activated, test.vim will return the correct "Not a test file" error.

However, your original problem was that the test files that you're running don't match the default test.vim's Mocha runner pattern. You can easily fix that by changing the file pattern (you can put it in your `.vimrc`):

```vim
[...]

PR #214, opened 2017-08-22
Fix detetion for Mocha and Jest

codeinabox:
Arises from #213

### DIFF
--- a/autoload/test/javascript/jest.vim
+++ b/autoload/test/javascript/jest.vim
@@ -4,7 +4,7 @@ endif
 
 function! test#javascript#jest#test_file(file) abort
   return a:file =~# g:test#javascript#jest#file_pattern
-	  \ && (test#javascript#has_package('jest') || !empty(test#javascript#jest#executable()))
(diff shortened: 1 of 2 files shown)


janko-m:
@codeinabox Can you explain the reasoning behind checking for `filereadable('node_modules/.bin/<runner>')`?

codeinabox:
@Akhetopnu could you try switching to this branch and let me know if it fixes your issue.

codeinabox:
@janko-m that was how the existing logic worked but given that would only be true if it's in package.json then it's probably redundant right?

janko-m:
@codeinabox Yes, I think we can remove it, and have the JavaScript runners match only when the corresponding testing framework is listed in `package.json` (the `test#javascript#has_package` condition).

janko-m:
@codeinabox Thanks! I also added https://github.com/janko-m/vim-test/commit/0877dfcace9f77f2bb5d67a21a2f510cfe1e2bd0 which applies that change for other JS test runners as well.

Issue #215, opened 2017-08-25
Read lines in a test file to determine runner


dstrunk:
Great plugin, I love being able to use it for Elixir and Rails work without worrying about different configurations. I'm now building an app in PHP's Laravel, and got started with their integration testing framework, [Dusk](https://laravel.com/docs/5.4/dusk). Tests using Dusk look so similar to PHPUnit that I'm having a hard time figuring out how to conditionally run dusk based on these three conditions:

1. Existence of `laravel/dusk` in the `composer.json` file
2. Folder structure: dusk files will exist in the `Browser/YourTest.php` folder
3. A `use` statement in the test file, `use Laravel\Dusk\Browser;`

Is there some way (via the exposed API, preferably) to perform a regex on the currently open file to find a line and conditionally use a different runner?

I know number 1 can be solved using something like [`s:rails_version()`](https://github.com/janko-m/vim-test/blob/453db3d6dce641f91c66731aad971530729ca172/autoload/test/ruby/rails.vim#L39-L55), and I believe number 2 can be solved using regex.

Interested to see if number 3 could be solved, or if the first two are fine for adding a dusk runner to this project?

codeinabox:
The third approach is very much possible, have a look what I did in https://github.com/janko-m/vim-test/blob/master/autoload/test/php/phpspec.vim

janko-m:
@dstrunk It's not 100% correct to do `search()`, because you can call `:Test<Command>` from application files (which test.vim detects and runs that command on the last touched test file instead). It would be better to use `readfile()` and `filter()`:

```vim
function! test#php#laravel#test_file(file) abort
  return a:file =~# g:test#php#laravel#file_pattern
    \ && !empty(filter(readfile(a:file), 'v:val =~# "use Laravel\\Dusk\\Browser"'))
readfile(file)
```

codeinabox:
@dstrunk let us know if you need any help implementing the test runner. If you create and test against some fixtures - see the examples in spec folder - it should be relatively straight forward.

dstrunk:
Thanks, appreciate the pointers! I'll work on this in the afternoon and hope to submit a PR this weekend.

PR #216, opened 2017-08-25
Change PhpSpec test file detection to use readfile

codeinabox:
Arises from #215. I am able to get it to work when searching for "PhpSpec" but not "PhpSpec\ObjectBehavior" for reason.

### DIFF
--- a/autoload/test/php/phpspec.vim
+++ b/autoload/test/php/phpspec.vim
@@ -3,9 +3,8 @@ if !exists('g:test#php#phpspec#file_pattern')
 endif
 
 function! test#php#phpspec#test_file(file) abort
-  if a:file =~# g:test#php#phpspec#file_pattern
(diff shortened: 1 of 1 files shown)


janko-m:
Thanks!

> I am able to get it to work when searching for "PhpSpec" but not "PhpSpec\ObjectBehavior" for reason.

You mean if the file contains content `PhpSpec\ObjectBehaviour`, you cannot detect it via this method? I tested it now, and you're right, the following doesn't work:

```vim
filter(readfile(a:file), 'v:val =~# "PhpSpec\\ObjectBehaviour"')
```

However, for some reason if we use single quotes, it works (you can insert single quotes inside single quotes by doubling them):

```vim
filter(readfile(a:file), 'v:val =~# ''PhpSpec\\ObjectBehaviour''')
```

PR #217, opened 2017-08-25
Added Laravel Dusk runner

dstrunk:
Add Laravel Dusk tests to vim-test.

Implementation should be almost exactly similar to PhpUnit; tests should only run differently if `use Laravel\Dusk\Browser` exists in the file.

To avoid worrying about folder structure (that seemed a bit brittle to me), I just tested for the presence of 'use Laravel\Dusk\Browser' and put the Dusk test runner before PHPUnit's to ensure it ran first (since it's using the same file matching pattern).

### DIFF
--- a/autoload/test/php/dusk.vim
+++ b/autoload/test/php/dusk.vim
@@ -1,0 +1,43 @@
+if !exists('g:test#php#dusk#file_pattern')
+  let g:test#php#dusk#file_pattern = '\v(t|T)est\.php$'
+endif
+
+function! test#php#dusk#test_file(file) abort
+  return a:file =~# g:test#php#dusk#file_pattern
+    \ && !empty(filter(readfile(a:file), 'v:val =~# ''use Laravel\\Dusk\\Browser'''))
+endfunction
+
+function! test#php#dusk#build_position(type, position) abort
+  if a:type == 'nearest'
+    let name = s:nearest_test(a:position)
+    if !empty(name) | let name = '--filter '.shellescape(name, 1) | endif
+    return [name, a:position['file']]
+  elseif a:type == 'file'
+    return [a:position['file']]
+  else
+    return []
+  endif
+endfunction
+
+function! test#php#dusk#build_args(args) abort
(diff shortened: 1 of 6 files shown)


autoload/test/php/dusk.vim
```diff
@@ -0,0 +1,43 @@
+if !exists('g:test#php#dusk#file_pattern')
+  let g:test#php#dusk#file_pattern = '\v(t|T)est.php$'
```
janko-m: Even though it probably wouldn't matter in the real world, it would be correct to escape the dot (`.`) here and write `\v(t|T)est\.php$`, as otherwise it means "any character".

.gitignore
```diff
@@ -1,3 +1,4 @@
 VimFlavor.lock
 .vim-flavor/
 /doc/tags
+.DS_Store
```
janko-m: This is Mac OS-specific, it should be put into a global gitignore.

autoload/test/php/dusk.vim
dstrunk: Good catch—I was wrestling with a folder-centric regex of this PR before deciding to just move the test runner to run before PHPUnit's—must've missed this in the shuffle.

I'll take care of these two reviews—would you prefer an amended commit, or a separate one that you can squash when merging?

janko-m: I'm good with an amended commit.

janko-m:
Let's wait for @codeinabox to take a look, and afterwards I'll merge.

dstrunk:
oop, good call, missed that … `doc/test.txt` added :)

doc/test.txt
```diff
@@ -82,6 +82,8 @@ In all commands [args] are forwarded to the underlying test runner.
 :M [args]                    Uses the `m` command.
 
                                                 *test-:Intern*
+PR
```
codeinabox: Typo

dstrunk: Sorry about that, overzealous commit

PR #221, opened 2017-09-18
Added a new command to tell the user whether the test exists

trisys3:
My main purpose with this is to automatically test on save, or perhaps on write, as `ale` does. I added an example `autocmd` to the docs file, though I haven't tested it yet.

If you want to go a different way with this, go ahead. This was more to give you an example of what I wanted to do, that you could refer to, rather than just opening another issue.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,6 +58,7 @@ nmap <silent> <leader>T :TestFile<CR>
 nmap <silent> <leader>a :TestSuite<CR>
 nmap <silent> <leader>l :TestLast<CR>
 nmap <silent> <leader>g :TestVisit<CR>
+nmap <silent> <leader>e :TestExists<CR>
 ```
 
 | Command          | Description                                                                                                                                                                                                                                                                            |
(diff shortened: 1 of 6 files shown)


trisys3:
I'm not sure why my tests failed Travis CI, they worked fine on my computer with a clean working directory. Can you point me in the right direction as to why?

janko-m:
@trisys3 Thank you for the PR! And sorry for the late response.

I agree this would be useful to have in vim-test, though I would prefer to only have the `test#exists()` function (good name, btw 👏). I think having Vim commands makes sense when we expect them to be run by users (since it's easier to run commands then calling functions), but this functionality is only meant to be used in scripts.

I'm also baffled to why the tests are suddenly failing, but it looks like it's related to the last commit. Perhaps `branch: "..."` specification isn't valid for a `VimFlavor` file, perhaps you could try without it.

trisys3:
As I said, I tried this on my computer and it worked fine. I had a clean working directory, and I have only been committing there on my computer, and no one else has push access on my repositories, unless GitHub itself got hacked, which would be a bigger problem. Maybe when I update the branch to remove the `TestExists` command as you suggested, Travis will sort itself out.

I hope I am not being too forward, but I also intended to abstract out the way the `test#run`command figures out whether the test exists. However, that also depends on whether or not there was a previous test run, for example `TestLast` or if you change files and run `TestFile` again. I couldn't abstract everything, so I just copied it.

trisys3:
OK, now I'm confused.

I removed the projectionist plugin as you suggested, and also after determining that that was the only change besides test & doc changes that I committed just before the test broke. I couldn't get the projectionist expectation to work, anyway, as it kept saying either "b:projectionist does not exist" or "g:projectionist_heuristics does not exist".

Anyway, long story short, now Travis still fails with "test#exists is not  function", even though identical tests work on my computer with no changes to the working directory, which theoretically means my repository is exactly the same as Travis's minus build artifacts, etc. Why would `test#exists` not be a function, anyway?

I'm sorry to be taking so much time on this. If you want to just forget about my PR, I won't mind at this point. Now that I know how easy it is to test that a test exists, I can just add it to my `vimrc` manually. Worst case, it could live in the README. Otherwise, I guess we should either accept that there will be failing tests or not have those tests and leave my `test#exists` function untested. Your call.

janko-m:
This has been merged.

PR #223, opened 2017-09-19
Fix tabpage ordering for neovim 'basic' strategy

### LINKED ISSUE #222: 'basic' strategy for Neovim doesn't return you to original tabpage
Scenario: I've got two tabpages open, with the first tabpage active. With the `basic` strategy, when I run `:TestFile` it opens a new tabpage *in between* the two that were already open. When I close that tabpage, Vim drops me into the next tabpage, leaving me on tabpage two. That's not where I started. Confusion ensues! 😕

nelstrom:
Fixes #222 

Inspired by this PR on fugitive: https://github.com/tpope/vim-fugitive/pull/784

### DIFF
--- a/autoload/test/strategy.vim
+++ b/autoload/test/strategy.vim
@@ -4,7 +4,7 @@ endfunction
 
 function! test#strategy#basic(cmd) abort
   if has('nvim')
-    tabnew
+    -tabnew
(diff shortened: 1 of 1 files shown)


janko-m:
Looks good, thank you!

PR #224, opened 2017-09-23
Add Dispatch! background variation

duleorlovic:
I see that we have [Dispatch](https://github.com/janko-m/vim-test/blob/master/autoload/test/strategy.vim#L35) strategy, but I would like to run test in background, so I added this dispatch_background strategy


janko-m:
@duleorlovic Thanks for the PR, looks good!

PR #227, opened 2017-10-09
Add regex to force only a single test to run

FatBoyXPC:
Sometimes, a test method is a bit too generic and you end running multiple tests instead of **just** the nearest test. Take the following methods for example: `testFoo`, `testFooBar`, `testFooBaz`. If you test nearest inside `testFoo` then you end up running all 3 of those methods. Since phpunit supports regex, this syntax prevents that from happening. I specifically ran into this when testing a fluent query builder that had a lot of `where` methods, such as `where`, `whereNull`, `whereNotNull`, etc.

I'd be more than willing to run a test to ensure that phpunit only runs 1 test (say by capturing the output of phpunit) I'd be down to write it, but I'm not familiar with vspec at all so I would need some direction!

### DIFF
--- a/autoload/test/php/phpunit.vim
+++ b/autoload/test/php/phpunit.vim
@@ -9,7 +9,7 @@ endfunction
 function! test#php#phpunit#build_position(type, position) abort
   if a:type == 'nearest'
     let name = s:nearest_test(a:position)
-    if !empty(name) | let name = '--filter '.shellescape(name, 1) | endif
+    if !empty(name) | let name = '--filter '.shellescape(name.'$', 1) | endif
     return [name, a:position['file']]
   elseif a:type == 'file'
     return [a:position['file']]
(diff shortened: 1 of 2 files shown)


FatBoyXPC:
Awesome! Again, just let me know if you want me to write any extra tests!

codeinabox:
These tests look fine to me - though I'd like to manually test the filter locally - but should we also add the starts with filter? https://stackoverflow.com/questions/10523700/how-do-i-do-starts-with-regex-filter-in-phpunit

FatBoyXPC:
Thanks for bringing that back up. I actually looked at doing starts with and ends with specifically for this but I was unable to get starts with to work when I tried it out. However, after playing with it a little bit more, I was able to get it working, commit incoming!

codeinabox:
Thanks again for this PR. I'll go over it throughly late next week when I am back from holiday.

FatBoyXPC:
Awesome, thanks!

FatBoyXPC:
Woo! Thanks!

PR #234, opened 2017-10-23
Add documentation on test#project_root to Vim help

codeinabox:


janko-m:
Great catch, thanks! 👍

Issue #232, opened 2017-10-23
pytest: nearest: picks up inline class


blueyed:
Given the following Python code:

```python
def test_foo():
    class CustomException(Exception):
        pass
    mocker.patch('some.module', side_effect=CustomException())

    assert something
```

`echo test#python#pytest#build_position('nearest', {'file': bufname('%'), 'line': line('.'), 'column': col('.')})` at the end of the function will result in `['t.py::CustomException']`.

It should use `test_foo` instead.

codeinabox:
Thanks for reporting this. I will try to look into this week but please free to raise a PR.

codeinabox:
I have recreated the issue. To confirm the issue only happens when it is run inside the test?

blueyed:
> To confirm the issue only happens when it is run inside the test?

Yes, somewhere below the `CustomException`.

codeinabox:
Closing as the PR has been merged in

PR #235, opened 2017-10-24
Check Python test class namespace is prefixed with Test

codeinabox:
This addresses the bug reported in #232

### DIFF
--- a/autoload/test/python.vim
+++ b/autoload/test/python.vim
@@ -1,4 +1,4 @@
 let test#python#patterns = {
   \ 'test':      ['\v^\s*%(async )?def (test_\w+)'],
-  \ 'namespace': ['\v^\s*class (\w+)'],
+  \ 'namespace': ['\v^\s*class (Test\w+)'],
 \}
--- a/spec/fixtures/nose/test_method.py
+++ b/spec/fixtures/nose/test_method.py
@@ -1,2 +1,9 @@
 def test_numbers():
     assert 1 == 1
+
+def test_foo():
+    class CustomException(Exception):
+        pass
+    mocker.patch('some.module', side_effect=CustomException())
+
+    assert something
--- a/spec/pytest_spec.vim
+++ b/spec/pytest_spec.vim
@@ -27,6 +27,11 @@ describe "PyTest"
     TestNearest
 
     Expect g:test#last_command == 'py.test test_method.py::test_numbers'
+
+    view +6 test_method.py
+    TestNearest
+
+    Expect g:test#last_command == 'py.test test_method.py::test_foo'
   end
 
   it "runs file test if nearest test couldn't be found"


codeinabox:
@janko-m trying to figure out why this is happening, shouldn't `g:test#python#patterns` ensure that the test name must start with `test_`?

janko-m:
@codeinabox The problem isn't that test.vim is not picking up the the `test_*` method, but that it's picking up the class that's inside the method, due to the `namespace` regex in `test#python#patterns`. 

I just took a look at "[Grouping multiple tests in a class](https://docs.pytest.org/en/latest/getting-started.html#grouping-multiple-tests-in-a-class)" and "[Conventions for Python test discovery](https://docs.pytest.org/en/latest/goodpractices.html#test-discovery)", and it seems that we're lucky, because names of classes that hold the `test_*` methods *need* to start with `Test`, so we just need to update the `namespace` regex with this constraint.

codeinabox:
@janko-m thank you for spotting the test class naming convention. Have updated the PR accordingly.

@blueyed could you please test this branch against your tests.

blueyed:
Will do.
However the Reeves should likely also handle unittest based tests: https://docs.pytest.org/en/latest/unittest.html

codeinabox:
Good point. We should make the fixtures we have for Python are representative for the majority of cases

autoload/test/python.vim
```diff
@@ -1,4 +1,4 @@
 let test#python#patterns = {
   \ 'test':      ['\v^\s*%(async )?def (test_\w+)'],
-  \ 'namespace': ['\v^\s*class (\w+)'],
+  \ 'namespace': ['\v^\s*class (Test\w+)'],
```
blueyed: I think something like `'\v^\s*class .*Test'` could work, i.e. a class with `Test` in the name, or derived from something with `Test` in the name.
To be more strict `'\v^\s*class (Test\w+|.*TestCase.*)'` could be used maybe.

janko-m: I agree with the stricter version 👍

codeinabox:
@blueyed question regarding subclass tests, given the following test

```python
import unittest
import pytest

@pytest.mark.usefixtures("db_class")
class MyTest(unittest.TestCase):
    def test_method1(self):
        assert hasattr(self, "db")
        assert 0, self.db   # fail for demo purposes

    def test_method2(self):
        assert 0, self.db   # fail for demo purposes
```

If I do test nearest on line 7, should the expected command be `py.test test_subclass.py::MyTest::test_method1`?

blueyed:
Yes.  Seems to work good.

blueyed:
Assuming line 7 is `        assert hasattr(self, "db")`.

janko-m:
Oops, I merged to soon, I didn't notice that we didn't add support for `unittest.TestCase` yet.

Issue #225, opened 2017-10-02
Feature request: Webdriver.io support


snoblenet:
Is it possible to add this please?

The CLI options are:

- Run all tests using the default configuration file: `wdio wdio.conf.js`
- Run a named suite of tests using the default configuration file: `wdio wdio.conf.js --suite NAME`
- Run just one test using the default configuration file: `wdio wdio.conf.js ---spec FILE_PATH_RELATIVE_TO_PROJECT_ROOT`

(Change `wdio` to `./node_modules/.bin/wdio` to use local wdio.)

More details: http://webdriver.io/guide/testrunner/organizesuite.html

codeinabox:
What file pattern do the tests have?

snoblenet:
If I understand the question correctly, the `wdio.conf.js` configuration file exports a location for the test files. Webdriver.io will search the top-level of that directory plus any sub-directories for test files. Webdriver.io outsources the actual test running to Mocha, Cucumber or Jasmine.

codeinabox:
Feel free to raise a PR or I could try picking it up when I've returned from holiday

snoblenet:
cheers -- I've taken to just running Webdriver.io tests in a separate terminal tab -- that's working for me for the time being

codeinabox:
I've made a start at adding support. Should be able to support TestFile and TestSuite. Will let you know when there is a branch you can test against.

snoblenet:
awesome

codeinabox:
Could you please try out the branch `feature/webdriver`

snoblenet:
will do thx

snoblenet:
Thanks for all your work on this.
I changed my `init.vim` to:
```
Plug 'janko-m/vim-test', { 'branch': 'feature/webdriver' }
let test#javascript#webdriver#file_pattern = '.js'
let g:test#runner_commands = ['Mocha', 'WebdriverIo']
```
I then CDd into the plug-in directory, noticed that `master` was still checked out, and pulled and checkout `feature/webdriver` instead manually.

I restarted Neovim, entered a spec file in a Webdriver.io project, and ran `:WebdriverIo`.

I got the message: `E:492 Not an editor command`.

By contrast, when I entered `:Mocha`, it attempted to run the file.

Inside the file, I then tried entering `:TestFile`, but I got a `Not a test file` error.

The full path to the file would be described by `/test/**/**/*.js`.

Any suggestions?

janko-m:
[...]

PR #236, opened 2017-10-24
Webdriver.io support

codeinabox:
This adds Webdriver.io support as requested in #225

### DIFF
--- a/autoload/test/javascript/webdriverio.vim
+++ b/autoload/test/javascript/webdriverio.vim
@@ -1,0 +1,29 @@
+if !exists('g:test#javascript#webdriverio#file_pattern')
+  let g:test#javascript#webdriverio#file_pattern = '\vtest/specs/.*\.js$'
+endif
+
+function! test#javascript#webdriverio#test_file(file) abort
+  return a:file =~# g:test#javascript#webdriverio#file_pattern
+    \ && test#javascript#has_package('webdriverio')
+    \ && !empty(glob('wdio.conf.js'))
+endfunction
+
+function! test#javascript#webdriverio#build_position(type, position) abort
+  if a:type == 'nearest' || a:type == 'file'
+    return ["--spec", a:position['file']]
+  else
+    return []
+  endif
+endfunction
+
+function! test#javascript#webdriverio#build_args(args) abort
+  return ['wdio.conf.js'] + a:args
+endfunction
+
+function! test#javascript#webdriverio#executable() abort
+  if filereadable('node_modules/.bin/wdio')
+    return 'node_modules/.bin/wdio'
+  else
+    return 'wdio'
+  endif
+endfunction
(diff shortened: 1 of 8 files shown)


autoload/test/javascript/webdriver.vim
```diff
@@ -0,0 +1,30 @@
+if !exists('g:test#javascript#webdriver#file_pattern')
+  let g:test#javascript#webdriver#file_pattern = '\vtest/specs/.*\.js$'
+endif
+
+function! test#javascript#webdriver#test_file(file) abort
+  return a:file =~# g:test#javascript#webdriver#file_pattern
```
janko-m: Should we also add a `\ && test#javascript#has_package('webdriverio')` here?

autoload/test/javascript/webdriver.vim
```diff
@@ -0,0 +1,30 @@
...
+    \ && !empty(glob('wdio.conf.js'))
+endfunction
+
+function! test#javascript#webdriver#build_position(type, position) abort
+  return []
+
+  if a:type == 'file'
+    return [a:position['file']]
```
janko-m: According to https://github.com/janko-m/vim-test/issues/225#issue-261966924 this should be 

```vim
return ["--spec", a:position['file']]
```

autoload/test/javascript/webdriver.vim
```diff
@@ -0,0 +1,30 @@
...
+  let g:test#javascript#webdriver#file_pattern = '\vtest/specs/.*\.js$'
+endif
+
+function! test#javascript#webdriver#test_file(file) abort
+  return a:file =~# g:test#javascript#webdriver#file_pattern
+    \ && !empty(glob('wdio.conf.js'))
+endfunction
+
+function! test#javascript#webdriver#build_position(type, position) abort
+  return []
```
janko-m: It seems that thins line wasn't removed after the function was extended below.

spec/webdriver_spec.vim
```diff
@@ -0,0 +1,28 @@
...
+    cd -
+  end
+
+  it "runs file tests"
+    view test/specs/example.js
+    TestFile
+
+    Expect g:test#last_command == 'wdio wdio.conf.js ---spec test/specs/example.js'
+  end
+
```
janko-m: Let's also add a test where we test that `:TestNearest` executes a file test (same as `:TestFile`).

codeinabox:
@janko-m given the command to run the tests is `wdio` should I call the runner that instead of `webdriverio` or better to be verbose?

janko-m:
@codeinabox I would prefer to use the full name for explicitness, the `wdio` CLI was abbreviated probably just for quicker typing in the shell.

autoload/test/javascript/webdriverio.vim
```diff
@@ -0,0 +1,29 @@
+if !exists('g:test#javascript#webdriverio#file_pattern')
+  let g:test#javascript#webdriverio#file_pattern = '\vtest/specs/.*\.js$'
```
janko-m: Actually, are tests really located in the `test/specs` directory? In any case, it would be better to have an `^` at the beginning of the regex.

PR #237, opened 2017-10-26
pytest: prefer pytest executable without dot

blueyed:


autoload/test/python/pytest.vim
```diff
@@ -7,7 +7,7 @@ function! test#python#pytest#test_file(file) abort
     if exists('g:test#python#runner')
       return g:test#python#runner == 'pytest'
     else
-      return executable('py.test')
+      return executable('pytest') || executable('py.test')
```
blueyed: Not sure if keeping the 2nd one makes sense?!

janko-m: If `pytest` will always exist alongside `py.test`, then by all means remove the `py.test` check.

janko-m:
I pushed your commit to master with tests added, and modified the logic to still support `py.test` when using versions 1.x and 2.x; as far as I could tell 3.x starts using the `pytest` executable name.

blueyed:
Thanks!

Issue #240, opened 2017-11-04
How to disable change directory when running test?


taufek:
I'm using vim-tmux-runner dispatch strategy (`test#strategy = "vtr"`) to run my test in tmux pane that runs a docker container. In the container, my project located on different path from the host machine. Whenever I run the tests, seems like `vim-test` will automatically change to my project directory according to my host machine which resulted as "no such file or directory" error as shown below.

<img width="645" alt="chdir" src="https://user-images.githubusercontent.com/10803354/32405873-b933f966-c1a8-11e7-9ebd-07867e9b3105.png">

Is there a way to disable that change directory that I highlighted in above screenshot?

codeinabox:
I know this was discussed in #183 if that helps

taufek:
Thanks @codeinabox .

I've created a PR to fix my issue. Thoughts anyone?

https://github.com/janko-m/vim-test/pull/241

codeinabox:
Thank you. @janko-m is probably the best one to comment on this.

codeinabox:
@janko-m what are your thoughts on this.

janko-m:
Sorry, I haven't been able to look this yet. At the moment I'm more inclined to making automatic `cd`-ing *opt-in* rather than opt-out, as it was added as a convenience in case you're using something like `vtr` and in the vim-test's tmux pane you wander off to some other directories.

janko-m:
Note that here I don't mean the `cd`-ing that happens on the test.vim level (in `autoload/test.vim`), I mean the `cd`-ing that happens on the strategy level (in `autoload/test/strategy.vim`).

janko-m:
@taufek Let me know if https://github.com/janko-m/vim-test/pull/259 resolved the issue.

taufek:
@janko-m , it does. Thanks for helping me to fix this. 👍

PR #242, opened 2017-11-04
[Option 1] Update Change Directory Logic When Running Tests

taufek:
I'm having an issue with my setup where I'm using `vtr` dispatch strategy and I'm targeting
a tmux pane that runs a docker container. My project path in the docker container is different from the one in my host machine. So whenever I'm executing a test, there is `cd getcwd()` that will
be called and it resulted with `no such file or directory`.

Here I'm moving the code related to changing directory based on defined project root to the strategy file.

I've added a function that determine which project directory to change to with following priority

1. project_root
2. cwd (current working directory)

### DIFF
--- a/autoload/test.vim
+++ b/autoload/test.vim
@@ -144,10 +144,6 @@ function! s:before_run() abort
   if &autowrite || &autowriteall
     silent! wall
   endif
-
-  if exists('g:test#project_root')
-    execute 'cd' g:test#project_root
-  endif
 endfunction
 
 function! s:after_run() abort
--- a/autoload/test/strategy.vim
+++ b/autoload/test/strategy.vim
@@ -125,7 +125,7 @@ endfunction
 
 function! s:pretty_command(cmd) abort
   let clear = !s:Windows() ? 'clear' : 'cls'
-  let cd = 'cd ' . shellescape(getcwd())
+  let cd = s:cd_command()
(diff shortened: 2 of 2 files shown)


aignas:
I like this. I was just thinking about something similar for #249.

taufek:
@janko-m ,

What is your thought on this change? Anything I can do to get this merge?

janko-m:
@taufek Thank you for the PR! What do you think about https://github.com/janko-m/vim-test/issues/240#issuecomment-343943388?

I seems that the `'cd' . getcwd()` being automatically executed is causing the problem, but that was just added as a convenience for when someone wonders off in their tmux pane to somewhere else, which is definitely an edge case and test.vim shouldn't be expected to handle this.

I think this change wouldn't be ideal, as `pretty_command` is not executed every time (e.g. when using the default `basic` strategy with Neovim). It's actually not supposed to have any side effects, I see that `'cd' . getcwd()` actually breaks this, so it shouldn't really be there.

taufek:
@janko-m ,

Thanks for the tip. I've created a new PR based on your suggestion. Let me know if I got it wrong.

https://github.com/janko-m/vim-test/pull/259

janko-m:
Closing in favour of https://github.com/janko-m/vim-test/pull/259.

Issue #243, opened 2017-11-08
Broken Python patterns for nearest tests


michaelmanganiello-eb:
Both #235 and #239 are based on some wrong assumptions, and has made vim-test to fail in some scenarios, at least for my particular usage:

1. We shouldn't have to change our existing code to prefix test classes with `Test`, when existing test classes with `Test` somewhere in their names already qualify for our test runners.
2. #239 forces us to use `import unittest`, and not even considers partial imports like `from unittest import TestCase`.
3. No support for test class inheritance, unless the child class is also prefixed with `Test`.

IMHO, these scenarios are more common than bug #232.

codeinabox:
Thank you for the bug report, could you please include an example of the test you are trying to run and we can fix the implementation against it.

michaelmanganiello-eb:
Using [unittest docs](https://docs.python.org/2/library/unittest.html) as an example, these are some failing scenarios:

```
import unittest

class SimpleWidgetTestCase(unittest.TestCase):
    pass

class DefaultWidgetSizeTestCase(SimpleWidgetTestCase):
    def test_widget(self):
        pass

# Should result in DefaultWidgetSizeTestCase:test_widget
```

```
from unittest import TestCase

class DefaultWidgetSizeTestCase(TestCase):
    def test_widget(self):
        pass

# Should result in DefaultWidgetSizeTestCase:test_widget
```

codeinabox:
I've created a branch to fix it, could you please give it a try.

codeinabox:
Please let know if you have any further feedback on #245 otherwise shall merge during the week.

codeinabox:
I've merged the fix into master. Could you please let me know if you still have issues.

janko-m:
I will close this, as https://github.com/janko-m/vim-test/pull/245 should have resolved the issue.

chrisnovak:
Looks like this is still an issue.

PR #245, opened 2017-11-10
Simplify the Python namespace regular expression

codeinabox:
To address #243

### DIFF
--- a/autoload/test/python.vim
+++ b/autoload/test/python.vim
@@ -1,4 +1,4 @@
 let test#python#patterns = {
   \ 'test':      ['\v^\s*%(async )?def (test_\w+)'],
-  \ 'namespace': ['\v^\s*class (Test\w+)|(\w+)\(unittest\.TestCase\)'],
+  \ 'namespace': ['\v^\s*class (\w+)'],
 \}
--- a/spec/fixtures/nose/test_class.py
+++ b/spec/fixtures/nose/test_class.py
@@ -13,3 +13,7 @@ class Test_underscores_and_123(Subclass):
 class UnittestClass(unittest.TestCase):
     def test_unittest(self):
         assert 1 == 1
+
+class SomeTest(TestCase):
+    def test_foo(self):
+        foo = date(2017, 11, 16)
--- a/spec/pytest_spec.vim
+++ b/spec/pytest_spec.vim
@@ -33,21 +33,33 @@ describe "PyTest"
 
     Expect g:test#last_command == 'pytest test_class.py::Test_underscores_and_123'
 
+    view +11 test_class.py
+    TestNearest
+
+    Expect g:test#last_command == 'pytest test_class.py::Test_underscores_and_123::test_underscores'
+
     view +13 test_class.py
     TestNearest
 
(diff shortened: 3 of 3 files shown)


mixxorz:
Broken on this.

```python
# tests/test_foo.py

class SomeTest(TestCase):

    def test_foo(self):
        foo = date(2017, 11, 16)
```

It resolves as `tests.test_foo.SomeTest.test_foo.date` when run from line 6.

codeinabox:
@mixxorz I've created a test to replicate but I get a different issue

michaelmanganiello-eb:
I'm not familiar with VimScript, but I'll try to be helpful.

I don't get the `'\v^\s*class (Test\w+)|(\w+)\(.*\)'` regex, because the pipe seems to be wrong. It's making `\v^\s*class (Test\w+)` and `(\w+)\(.*\)` two different options. The second one will match to any function call or function definition.

Maybe what we really want is `'\v^\s*class (Test\w+|\w+)\(.*\)'` (look at the difference in the first matching group). However, that could be simplified to `'\v^\s*class (\w+)\(.*\)'`, and, as far as I understand, should make #232 to happen again.

IMHO the real reason for #232 is that we are matching for namespaces even before we already have a match for the test name. For example:

```
1    class TestFoo(BaseTests):
2        def test_foo():
3            class CustomException(Exception):
4                pass
5            assert 1 == 1
```

Should start looking for the namespace *after* the test name was found (line 2).

Of course, we also want to be able to test a whole namespace if we want (e.g. running all the `TestFoo` tests). So I would think the best solution would be to keep this regex as it was before #235 (`'\v^\s*class (\w+)'`), and clean the namespace when we find a test for the first time. That means, setting `namespace` as an empty list here: https://github.com/janko-m/vim-test/blob/612c75eb60948d048122d1f5d0411a59a578a932/autoload/test/base.vim#L82, as we shouldn't have a namespace inside the test. Does that make sense?

codeinabox:
I've managed to get this working for all classes except not matching `class CustomException(Exception)`. @blueyed I know this is one we tried to support but handling this scenario seems tricky without making regular expressions more clever.

autoload/test/python.vim
```diff
@@ -1,4 +1,4 @@
 let test#python#patterns = {
   \ 'test':      ['\v^\s*%(async )?def (test_\w+)'],
-  \ 'namespace': ['\v^\s*class (Test\w+)|(\w+)\(unittest\.TestCase\)'],
+  \ 'namespace': ['\v^\s*class (Test\w+|\w+)'],
```
janko-m: Note that this regex is equivalent to `'\v^\s*class (\w+)'`.

codeinabox: Given the issues, do you think we should just simplify the regex down to that and treat classes defined instead a test block as more an edge case?

blueyed:
> clean the namespace when we find a test for the first time
> … as we shouldn't have a namespace inside the test. Does that make sense?

That would solve it, wouldn't it?

codeinabox:
@blueyed yes that would solve it but I'd need to check with @janko-m or review the existing code to find if that's possible.

In the interim it's probably best we simplify the regular expression with this PR so it works for the majority of cases.

Issue #246, opened 2017-11-12
Elixir ExUnit: Not a Test File


randre03:
Sorry in advance, I know you receive these questions all of the time, but I simply cannot figure out how to get vim-test to run my test suite when I run the command in a buffer that is not a test file. I am using Elixir and ExUnit. All my tests are in the following format (is this the correct regex???):
` let test#elixir#exunit#file_pattern = '*_test\.exs'`

I am using vim-rooter[https://github.com/airblade/vim-rooter](url) so that I can make fzf search in my project root (and not just the pwd). I thought it might be interfering with vim-test's need to know the pwd, but I removed it and still had the issue.

Thanks in advance - vim-test really is fantastic!

codeinabox:
What is the name of the file you are trying to run tests for?

randre03:
That would help, wouldn’t it?

`physics_test.exs`

- Roger

On Nov 12, 2017, 9:36 AM -0500, Mathew Attlee <notifications@github.com>, wrote:

What is the name of the file you are trying to run tests for?

codeinabox:
That should work. Are you running `TestFile` when that has focus on the buffer, or are you trying to run it from `physics.exs`. If the latter you need to set up projectionist alternate mappings, I've added this to the documentation #247

janko-m:
Note that vim-test uses regexes, not globs, so `'*_test\.exs'` won't work because that's not a valid regex, you would need to use `'.*_test\.exs'`. Also, note that by default vim-test uses its own regex mechanism which is similar but in many cases different than the POSIX standard, the POSIX standard can be turned on via `\v`. Note that the pattern also doesn't have to match the entire file, so you can remove the `.*`, and you should also anchor the end. With all this in mind, the following should work:

```vim
let test#elixir#exunit#file_pattern = '\v_test\.exs$'
```

codeinabox:
@janko-m should we update the default pattern? I assumed it was fine because it passes the specs

janko-m:
@codeinabox Yes, I think we should add `\v` to all patterns, because I cannot think of any reason we would want the regexes to behave in the way we're all used to 😃

janko-m:
@codeinabox Node that in Vim regexes, characters `.` and `$` still behave the same as in POSIX regexes, so that's why it works. But for example `(` characters are taken literally in Vim regexes, while `\(` would be the special character which creates a regex group; in POSIX standard it's the opposite.

codeinabox:
[...]

PR #248, opened 2017-11-13
Update all file patterns to be POSIX standard

codeinabox:
Arises from #246


jannikbecher:
Why should be the exunit pattern be "v_test.exs" instead of just "_test.exs"?

PR #249, opened 2017-11-18
[RFC] dotnet-xunit support

aignas:
This is initial support for the `dotnet-xunit` tool provided by the XUnit framework for C#.

### DIFF
--- a/autoload/test/csharp/xunit.vim
+++ b/autoload/test/csharp/xunit.vim
@@ -1,0 +1,72 @@
+if !exists('g:test#csharp#xunit#file_pattern')
+  let g:test#csharp#xunit#file_pattern = '\v\.cs$'
+endif
+
+function! test#csharp#xunit#test_file(file) abort
+  if fnamemodify(a:file, ':t') =~# g:test#csharp#xunit#file_pattern
+    if exists('g:test#csharp#runner')
+      return g:test#csharp#runner ==# 'xunit'
+    else
+      return s:is_using_dotnet_xunit_cli(a:file)
+          \ && (search('using Xunit', 'n') > 0)
+          \ && (search('[(Fact|Theory).*]', 'n') > 0)
+    endif
+  endif
+endfunction
+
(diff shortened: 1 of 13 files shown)


aignas:
I added a few spec tests and I think that for the initial implementatino it may be enough. The problems arise when trying to handle some cases when the design of the test classes is not ideal, but I guess we can think about it at some later point.

Could someone review this and let me know what else I should add in order to get this merged? :) Thanks :)

autoload/test/csharp/xunit.vim
```diff
@@ -0,0 +1,73 @@
...
+endfunction
+
+function! test#csharp#xunit#build_position(type, position) abort
+  " TODO: if we can hook into the before_run function, then in theory we
+  " could set the project_root dynamically before each run, which would
+  " remove the necessity of setting the g:test#project_root manually.
+  "
+  "let file = a:position['file']
+  "let project_path = test#csharp#get_project_path(file)
+  "let g:test#project_root = fnamemodify(project_path, ':p:h')
```
codeinabox: Does this commented out code need to remain?

codeinabox:
Thank you so much for your contribution, very much appreciated. Could you also update doc/test.txt with the test command please.

autoload/test/csharp/xunit.vim
aignas: Not necessarily. At the moment there is a bug, that the pwd needs to be the same as the solution file and if you have sub-folders in the project and autochdir enabled, then it will not work unless you manually change the folder to the project root. I thought that setting it dynamically would work, but that does not work and it needs the changes that are introduced in #242 or something similar.

Shall I remove this commented out code and create an issue?

codeinabox: Yes I think that would be better as commented out code can be a source of confusion. If you do create an issue you can add in the link in the comment block.

aignas:
@codeinabox, I rebased and squashed on top of the latest master and created an issue.

PR #255, opened 2017-12-11
Do Proper Checking for Tape

jalcine:
I work on a project that has a sub-dependency that has `tape` as a normal dependency vs a development dependency; thus making the original checker for `tape` think that it should use it over the top-most testing tool (that being mocha).

In other words, this avoids the blunt checking for testing tools and checks the the top-most project's approach.

This aims to use the same kind of detection that mocha does to counteract mis-defined projects like this.

---
This change is [Reviewable](https://reviewable.io/reviews/janko-m/vim-test/255)


janko-m:
Thanks, that change is definitely in line with other JavaScript runners that test.vim already has.

PR #258, opened 2017-12-15
Javascript Ava runner support

codeinabox:
This picks up the work done in #148 and updates it to use `has_package`

### DIFF
--- a/autoload/test/javascript/ava.vim
+++ b/autoload/test/javascript/ava.vim
@@ -3,13 +3,8 @@ if !exists('g:test#javascript#ava#file_pattern')
 endif
 
 function! test#javascript#ava#test_file(file) abort
-  if a:file =~# g:test#javascript#ava#file_pattern
-    if exists('g:test#javascript#runner')
-      return g:test#javascript#runner == 'ava'
-    else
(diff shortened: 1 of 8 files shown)


autoload/test/javascript/ava.vim
```diff
@@ -0,0 +1,46 @@
...
+  endif
+endfunction
+
+function! test#javascript#ava#build_args(args) abort
+  let args = a:args
+
+  if test#base#no_colors()
+    let args = ['--no-colors'] + args
+    let args = args + ['|', 'sed -e "s/
//g"']
```
janko-m: This was copied from Mocha, I would check whether this is needed (as it was specific to Mocha) and/or whether the `ava` runner even supports a `--no-colors` option.

codeinabox:
@tylucaskelley feel free to give this branch a spin, any feedback on the Ava support would be appreciated.

autoload/test/javascript/ava.vim
codeinabox: The option is slightly different https://github.com/avajs/ava#cli

I will update PR accordingly

PR #259, opened 2017-12-20
[Option 3] Remove Change Directory in Strategy

taufek:
**Why we need this change?**

I'm having an issue with my setup where I'm using vtr dispatch strategy and I'm targeting
a tmux pane that runs a docker container. My project path in the docker container is different from the one in my host machine. So whenever I'm executing a test, there is cd getcwd() that will
be called and it resulted with no such file or directory.

**What has changed?**

I've removed the related code for changing to test directory in strategy file.

### DIFF
--- a/autoload/test/strategy.vim
+++ b/autoload/test/strategy.vim
@@ -125,22 +125,20 @@ endfunction
 
 function! s:pretty_command(cmd) abort
   let clear = !s:Windows() ? 'clear' : 'cls'
-  let cd = 'cd ' . shellescape(getcwd())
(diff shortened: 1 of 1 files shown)


turino:
@janko-m I'm using `test#strategy = "terminal"`, which was relying on the automatic `cd`ing. What do I need to put in my vimrc to get vim-test to `cd` for me again?

janko-m:
@turino I just added back automatic `cd`-ing to `terminal` and `iterm` strategies, as I realized this default makes sense for those two strategies.

PR #261, opened 2018-01-02
Fixed phpunit filter

przepompownia:
Without this change no test was executed by phpunit 6.3.0.

See https://phpunit.de/manual/6.5/en/textui.html#textui.examples.filter-patterns

### DIFF
--- a/autoload/test/php/phpunit.vim
+++ b/autoload/test/php/phpunit.vim
@@ -9,7 +9,7 @@ endfunction
 function! test#php#phpunit#build_position(type, position) abort
   if a:type ==# 'nearest'
     let name = s:nearest_test(a:position)
-    if !empty(name) | let name = '--filter '.shellescape('::'.name.'$', 1) | endif
+    if !empty(name) | let name = '--filter '.shellescape('/::'.name.' .*$/', 1) | endif
     return [name, a:position['file']]
   elseif a:type ==# 'file'
     return [a:position['file']]
--- a/spec/phpunit_spec.vim
+++ b/spec/phpunit_spec.vim
@@ -27,29 +27,29 @@ describe "PHPUnit"
     view +9 NormalTest.php
     TestNearest
 
-    Expect g:test#last_command == "phpunit --colors --filter '::testShouldAddTwoNumbers$' NormalTest.php"
+    Expect g:test#last_command == "phpunit --colors --filter '/::testShouldAddTwoNumbers .*$/' NormalTest.php"
 
     view +14 NormalTest.php
     TestNearest
 
-    Expect g:test#last_command == "phpunit --colors --filter '::testShouldSubtractTwoNumbers$' NormalTest.php"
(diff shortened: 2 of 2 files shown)


codeinabox:
Thank you so much for this. Could you add a test for it please, should just be a case of updating the spec and adding a fixture if need be

przepompownia:
```
 $ vim-flavor help 
/usr/local/bin/vim-flavor:22:in `load': cannot load such file -- /var/lib/gems/2.3.0/gems/vim-flavor-2.2.1/bin/vim-flavor (LoadError)
         from /usr/local/bin/vim-flavor:22:in `<main>'
```
I have a problem with running `vim-flavor` on my local machine, but we have Travis.

harmenjanssen:
I can confirm the bug as well as the solution for PHPunit `3.7.37`!

Would love to have this solution merged into `master`.

codeinabox:
Great thank you for your contribution and patience.

harmenjanssen:
I'm sorry, my confirmation was a bit premature, it does _not_ work in phpunit `3.7`. 
Honestly `3.7` is _really_ old. I can understand if you would deprecate support for it. 

I'll see if I can upgrade on my end. My projects with current phpunit versions all work fine with the fix.

harmenjanssen:
I upgraded to phpunit 6.5.5 and can unfortunately confirm it's also not working there.
The line should be: 

```
if !empty(name) | let name = '--filter '.shellescape('/::'.name.'$/', 1) | endif
```

instead of 

```
if !empty(name) | let name = '--filter '.shellescape('/::'.name.' .*$/', 1) | endif
```

@przepompownia Can you confirm the above also works for phpunit 6.3? 
If so, I can offer a pull request with a complete fix later today.

janko-m:
What is the string that the pattern matches? What is the rest of the string after the test name? With the `.*`, executing nearest on `function testWorks()` would then also match tests which have the same prefix, e.g. `function testWorksWithVariable()`.

harmenjanssen:
You can provide "named data" (see slightly above this section: https://phpunit.de/manual/6.5/en/textui.html#textui.examples.filter-patterns).

But I don't think it matters much in the context of `vim-test` because it will always be the same value? I think it's safe to just omit it, since there's no meaningful data to supply it with. 
I don't have experience with it myself, maybe @przepompownia can elaborate a little.

codeinabox:
Is there an argument to rollback this change for now or is there a fix we can create quickly?

harmenjanssen:
I think it even works in phpunit 3.7 when you leave out the ` .*` part, so that's probably the quickest win 🙂

codeinabox:
Instead of the wild card don't we want a filter like this?

`--filter 'TestNamespace\\TestCaseClass::testMethod'`

harmenjanssen:
Ah, I guess you're right, that makes sense.

przepompownia:
Sorry for long silence. I noticed that my PR #265 fixed running tests with data providers but it made a regression for tests without data provider. @codeinabox you are right. The dollar sign was the only wrong element of filter that does not use regular expression.

Please see examples:
https://travis-ci.org/przepompownia/phpunit-filter-example

PR #265, opened 2018-01-15
Fix phpunit filter for tests with and without data provider

przepompownia:
https://travis-ci.org/przepompownia/phpunit-filter-example


janko-m:
@przepompownia Thanks for the quick patch!

FatBoyXPC:
This totally broke #227, I'll probably submit a new PR to make this work with data providers.

Issue #263, opened 2018-01-11
Error when running vim-flavor


codeinabox:
I am trying to run the tests locally but I get the following error:-

> gem install vim-flavor
> Successfully installed vim-flavor-2.2.1
> Parsing documentation for vim-flavor-2.2.1
> Done installing documentation for vim-flavor after 0 seconds
> 1 gem installed
> 
> vim-flavor test spec/
> -------- Preparing dependencies
> Checking versions...
>   Use junegunn/vader.vim ... 0.3.0
>   Use kana/vim-vspec ... 1.6.1
>   Use tpope/vim-fireplace ... v1.1
> Deploying plugins...
>   junegunn/vader.vim 0.3.0 ... skipped (already deployed)
>   kana/vim-vspec 1.6.1 ... skipped (already deployed)
>   tpope/vim-fireplace v1.1 ... skipped (already deployed)
> Completed.
> -------- Testing a Vim plugin
> spec/ava_spec.vim ............
>
> ~/workspace/vim-test/spec/support/helpers.vim, line 7
> Vim(function):E746: Function name does not match script file name: test#strategy#basic
> Bailout called.  Further testing stopped:  Unexpected error happened while processing a test script.
> FAILED--Further testing stopped: Unexpected error happened while processing a test script.

janko-m:
@codeinabox Which versions of Ruby, vim-flavor and Vim are you using?

codeinabox:
ruby 2.5.0p0 (2017-12-25 revision 61468) [x86_64-darwin17]
vim-flavor 2.2.1
I have NVIM v0.2.2 aliased to `vi` and Vi IMproved 8.0.1400 aliases to `vim`

janko-m:
@codeinabox I couldn't reproduce this, I've made the exact same setup as you have. I think we might have slightly different Vim versions, because the error that occurred is a Vim error. This is my version (installed with Homebrew):

<details>
<summary>vim --version</summary>

```
VIM - Vi IMproved 8.0 (2016 Sep 12, compiled Jul 26 2017 19:10:24)
Included patches: 1-503, 505-642
Compiled by root@apple.com
Normal version without GUI.  Features included (+) or not (-):
+acl             +file_in_path    -mouse_sgr       +tag_old_static
-arabic          +find_in_path    -mouse_sysmouse  -tag_any_white
+autocmd         +float           -mouse_urxvt     -tcl
-balloon_eval    +folding         +mouse_xterm     -termguicolors
[...]

PR #267, opened 2018-01-27
Move test strategies to the file with a proper name

voronkovich:
This PR fixes (I hope!) #263.

`From :h autoload`:
```
Using an autoload script 
                                                        autoload E746
This is introduced in the user manual, section 41.15.

Using a script in the "autoload" directory is simpler, but requires using
exactly the right file name.  A function that can be autoloaded has a name
like this: 

        :call filename#funcname()

When such a function is called, and it is not defined yet, Vim will search the
"autoload" directories in 'runtimepath' for a script file called
"filename.vim".  For example "~/.vim/autoload/filename.vim".  That file should
then define the function like this: 

        function filename#funcname()
           echo "Done!"
        endfunction

The file name and the name used before the # in the function must match
exactly, and the defined function must have the name exactly as it will be
called.
```

### DIFF
--- a/spec/nose_spec.vim
+++ b/spec/nose_spec.vim
@@ -1,9 +1,5 @@
 source spec/support/helpers.vim
 
-function! test#shell(cmd, strategy) abort
-  let g:test#last_command = substitute(a:cmd, ' --doctest-tests', '', '')
-endfunction
-
 describe "Nose"
 
   before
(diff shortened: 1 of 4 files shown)


codeinabox:
Thank for the quick update, now I am only getting the following errors:-

> 
> spec/strategy_spec.vim ....... 1/? 
> not ok 2 - strategy can be set via test#strategy as built-in
> function <SNR>1_bootstrap[24]..vspec#test[6]..<SNR>2_run_suites[8]..12[3]..test#execute[22]..test#shell[19]..test#strategy#neovim, line 2
> Vim(call):E117: Unknown function: termopen
> not ok 4 - strategy can be set per command
> function <SNR>1_bootstrap[24]..vspec#test[6]..<SNR>2_run_suites[8]..14[3]..test#execute[22]..test#shell[19]..test#strategy#neovim, line 2
> Vim(call):E117: Unknown function: termopen
> not ok 5 - strategy remembers strategy passed when running last test
> function <SNR>1_bootstrap[24]..vspec#test[6]..<SNR>2_run_suites[8]..15[4]..test#run_last[13]..test#shell[19]..test#strategy#neovim, line 2
> Vim(call):E117: Unknown function: termopen
> not ok 6 - strategy can be set for different granularities
> function <SNR>1_bootstrap[24]..vspec#test[6]..<SNR>2_run_suites[8]..16[5]..test#run[24]..test#execute[22]..test#shell[19]..test#strategy#neovim, line 2
> Vim(call):E117: Unknown function: termopen
> spec/strategy_spec.vim ....... Failed 4/7 subtests

voronkovich:
@codeinabox, It seems that it uses original strategy functions instead of ones defined in `spec/test/strategy.vim`. I don't know how to fix this problem :(.

voronkovich:
@codeinabox, I've fixed the problem by sourcing fake test strategies in the `before` block.

voronkovich:
@codeinabox, I think we do. Deleted function `test#shell` just removes the `--doctest-tests` option from output. IMHO, the tests is more readable (and more understandable) without it.
```viml
function! test#shell(cmd, strategy) abort
    let g:test#last_command = substitute(a:cmd, ' --doctest-tests', '', '')
endfunction
```

PR #268, opened 2018-01-27
Support running the nearest test for the PhpSpec

voronkovich:
PHPSpec allows to specify a line number to run only the specification defined on that line. See http://www.phpspec.net/en/stable/cookbook/console.html#run-command. So, it's better than nothing :).


janko-m:
Great, thanks for the patch!

PR #269, opened 2018-01-29
Add support for vim-themis

alexgenco:
This adds the [`themis`](https://github.com/thinca/vim-themis) runner. I've punted on adding `nearest` support because it will be a little complicated, seeing as it will have to convert a line number to the surrounding themis test function name. For now it just runs the whole file for both `file` and `nearest`.

### DIFF
--- a/autoload/test/viml/themis.vim
+++ b/autoload/test/viml/themis.vim
@@ -1,0 +1,28 @@
+if !exists('g:test#viml#themis#file_pattern')
+  let g:test#viml#themis#file_pattern = '\.vim$'
+endif
+
+if !exists('g:test#viml#themis#executable')
+  let g:test#viml#themis#executable = 'themis'
+endif
+
+function! test#viml#themis#test_file(file) abort
+  return a:file =~# g:test#viml#themis#file_pattern
+        \ && !empty(filter(readfile(a:file), 'v:val =~# ''\<themis#suite\s*('''))
+endfunction
+
+function! test#viml#themis#build_position(type, position) abort
+  if a:type == 'nearest' || a:type == 'file'
+    return [a:position['file']]
+  else
+    return []
+  endif
+endfunction
+
+function! test#viml#themis#build_args(args) abort
+  return a:args
+endfunction
+
+function! test#viml#themis#executable() abort
+  throw 'The `g:test#viml#themis#executable` variable needs to be set with the path to the "themis" executable'
+endfunction
--- a/spec/themis_spec.vim
+++ b/spec/themis_spec.vim
@@ -1,0 +1,45 @@
+source spec/support/helpers.vim
+runtime! plugin/vim-themis
+
+describe "Themis"
+
+  before
+    let g:test#viml#themis#executable = "/path/to/themis"
+    cd spec/fixtures/themis
+  end
+
+  after
+    call Teardown()
+    cd -
+  end
+
+  it "runs file test for nearest"
+    view +2 math.vim
+    TestNearest
+    messages
+
+    Expect g:test#last_command == "/path/to/themis math.vim"
+  end
+
+  it "runs file tests"
+    view +2 math.vim
+    TestFile
+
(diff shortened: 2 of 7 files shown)


alexgenco:
Updated the docs, and added Vader since it doesn't seem to be there.

autoload/test/viml/themis.vim
```diff
@@ -0,0 +1,38 @@
...
+    return []
+  endif
+endfunction
+
+function! test#viml#themis#build_args(args) abort
+  let args = a:args
+
+  if exists('test#viml#themis#args')
+    let args = g:test#viml#themis#args + args
+  endif
```
janko-m: This shouldn't be necessary, because you can already configure CLI arguments for any runner via `g:test#<language>#<runner>#options`, see the [docs](https://github.com/janko-m/vim-test#cli-options).

autoload/test/viml/themis.vim
```diff
@@ -0,0 +1,38 @@
+if !exists('g:test#viml#themis#file_pattern')
+  let g:test#viml#themis#file_pattern = '.vim$'
```
janko-m: The `.` seems to be a special even for Vim regexes, so you'll need to escape it.

autoload/test/viml/themis.vim
```diff
@@ -0,0 +1,38 @@
...
+
+  if exists('test#viml#themis#args')
+    let args = g:test#viml#themis#args + args
+  endif
+
+  return args
+endfunction
+
+function! test#viml#themis#executable() abort
+  if !executable(g:test#viml#themis#executable)
```
janko-m: I think this might throw an exception if `g:test#viml#themis#executable` is undefined. You can actually remove all of the code except the `throw`, because when `g:test#{language}#{runner}#executable` is defined, then the corresponding `g:test#{language}#{runner}#executable()` function is not called.

Since we won't be actually testing whether what's assigned is executable, we could rephrase the error message to be more straightforward:

```vim
function! test#viml#themis#executable() abort
  throw 'The `g:test#viml#themis#executable` variable needs to be set with the path to the "themis" executable'
endfunction
```

spec/themis_spec.vim
```diff
@@ -0,0 +1,48 @@
+source spec/support/helpers.vim
+runtime! plugin/vim-themis
+
+describe "Themis"
+
+  before
+    let g:test#viml#themis#executable =
+          \ getcwd() . '/.vim-flavor/deps/thinca_vim-themis/bin/themis'
```
janko-m: Since the `g:test#viml#themis#executable()` function will not get executed, we can actually assign this to whatever we want. So let's assign this to just `let g:test#viml#themis#executable = 'themis'`.

Then in tests we can replace assertions

```vim
Expect g:test#last_command == g:test#viml#themis#executable . ' math.vim'
```

with

```vim
Expect g:test#last_command == 'themis math.vim'
```

autoload/test/viml/themis.vim
alexgenco: Removing this makes this function just return `a:args` unchanged, do we need to define the function at all in this case?

janko-m: At the moment we still need to define it even if it just has `return a:args`.

autoload/test/viml/themis.vim
alexgenco: I made this change but the tests are hitting the function and throwing. It seems like maybe it's calling the function but throwing away the return value if the executable variable is set? Changing this to `return 'themis'` seems to work as expected.

janko-m: Oops, the function was still being executed (even though it's not used), I pushed a change now to `master` which doesn't call the function if the variable is defined.

alexgenco:
Ok I rebased onto master and squashed in changes to address you feedback.

alexgenco:
Sorry, I should have tried this earlier but it looks like the vspec runner trumps the themis runner, I'll open another PR, I guess to put the themis runner first.

PR #270, opened 2018-01-29
Put Themis before VSpec in runner ordering

alexgenco:
The VSpec `test_file` function trumps the Themis one, since they both
check the `.vim` extension. This puts `Themis` first, since it is a more
specific check.

This is a quick followup to https://github.com/janko-m/vim-test/pull/269. Sorry for the noise.


janko-m:
No problem, good call! 👍

Issue #209, opened 2017-08-04
Typescript Support


akinsho:
Hi, thanks for the excellent plugin which is amazing and works  really well with my javascript,

I'm currently working a lot more with typescript and have been playing with tweaking the javascript runner to work with typescript with no success. As far as I can tell what would be needed for this plugin to work with typescript would be to have the same set of runners as for JS but pass an argument somehow to point vim test to the typescript compiler's output directory. To run the tests on the JS that it compiles. I've tried creating a runner by duplicating the JS ones (which might be the wrong track) but I can't figure out how to point the plugin at a directory which is a variable or how it might recognise a TS file without a new runner for typescript. 

I'd love some advice and am happy to put in some leg work myself to add this runner.

codeinabox:
That would be great to add Typescript support. Could you please link to some documentation for the runner you would like to add and I'll advise as best as I can.

Akin909:
Hi, Thanks for the speedy response, what I sort of envisioned was for a typescript runner or flag to basically just run the javascript runner so the ts one would just redirect vim-test to the a directory a user could specify because all the runners for javascript are valid for typescript as far as I know, so rather than having to constantly update the TS runners and JS runners separately if the JS runners got updated then you could just use the new js one. Not that I know how that can be done 😨 , the runners I had in mind were at least [jest](https://github.com/facebook/jest/tree/master/examples/typescript) and [mocha](https://mochajs.org/). To use these in a normal ts project you would usually just configure them in the package json so stuff like calling the compiler could be left to those libraries maybe

codeinabox:
Perhaps the simplest solution might be a strategy

codeinabox:
To clarify are your tests written in JavaScript or TypeScript?

Akin909:
Tests are written in typescript which is compiled to javascript by the ts compiler, both jest and mocha run those. As for using a strategy Im not sure how to trigger a strategy for a file type that the plugin doesn't recognise, if I could invoke the plugin in a typescript file then if it runs mocha or jest commands they should both be able to manage the rest.

codeinabox:
I see your point. Given the way the plugin works, you would have to duplicate the existing JavaScript runners to make TypeScript versions. I don't think this can be avoided as they have different file types and the runner detection is based on that. @janko-m what are your thoughts?

[...]

PR #279, opened 2018-03-01
Support TypeScript for common JavaScript test frameworks

ajcrites:
This partially addresses #209 

Many test frameworks have their own transpile step and use the `.ts`/`.tsx` files as inputs.


codeinabox:
Thank you for this PR, very much appreciated. Do these frameworks support transpiling out of the box or is it an additional extension?

ajcrites:
Actually looking at this now I realize that it's pretty easy to update the pattern as part of your own vim configuration so these changes aren't needed. Sorry!

janko-m:
@ajcrites It would be nice that people using TypeScript don't need any additional configuration 😉

PR #280, opened 2018-03-01
Improve argument order for jest named testing

ajcrites:
The original order would run something like

```
jest $FILE -t $NAME
```

With custom options, this can cause issues. For example:

```
# let g:test#javascript#jest#options = '--reporters some-reporter'
jest --reporters some-reporter $FILE
```

jest attempt to use a reporter that matches the test filename.

This change uses the `--` before the input files to make them work
properly with custom and derived arguments.

```
jest --reporters some-reporter -t $NAME -- $FILE
```

### DIFF
--- a/spec/jest_spec.vim
+++ b/spec/jest_spec.vim
@@ -16,68 +16,68 @@ describe "Jest"
       view +1 __tests__/normal-test.js
       TestNearest
 
-      Expect g:test#last_command == 'jest __tests__/normal-test.js -t ''^Math'''
+      Expect g:test#last_command == 'jest -t ''^Math'' -- __tests__/normal-test.js'
 
       view +2 __tests__/normal-test.js
       TestNearest
 
-      Expect g:test#last_command == 'jest __tests__/normal-test.js -t ''^Math Addition'''
+      Expect g:test#last_command == 'jest -t ''^Math Addition'' -- __tests__/normal-test.js'
 
       view +3 __tests__/normal-test.js
       TestNearest
 
-      Expect g:test#last_command == 'jest __tests__/normal-test.js -t ''^Math Addition adds two numbers$'''
+      Expect g:test#last_command == 'jest -t ''^Math Addition adds two numbers$'' -- __tests__/normal-test.js'
     end
 
     it "aliases context to describe"
       view +1 __tests__/context-test.js
       TestNearest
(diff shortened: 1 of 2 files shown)


spec/jest_spec.vim
```diff
@@ -16,68 +16,68 @@ describe "Jest"
       view +1 __tests__/normal-test.js
       TestNearest
 
-      Expect g:test#last_command == 'jest __tests__/normal-test.js -t ''^Math'''
+      Expect g:test#last_command == 'jest -t ''^Math'' -- __tests__/normal-test.js'
```
codeinabox: Could you please elaborate on the `--` before the test name. I had a look at [the documentation](https://facebook.github.io/jest/docs/en/cli.html) and I can't find an example of this

spec/jest_spec.vim
```diff
@@ -107,7 +107,7 @@ describe "Jest"
     view outside-test.js
     TestFile
 
-    Expect g:test#last_command == 'jest outside-test.js'
+    Expect g:test#last_command == 'jest -- outside-test.js'
```
codeinabox: In this case wouldn't we just want it to be `jest outside-test.js` as per the official documentation?

ajcrites:
@codeinabox responding to your comments; I'm happy to explain. The `--` is a shell convention that signals to a command that you are no longer sending options to the command and it should treat all of the remaining words as arguments.

So for example if you wanted to search a file for `-i` you could try `grep -i $FILE`, but this won't work since `grep` thinks you mean to use `-i` as the ignore-case option. Instead, you can write `grep -- -i $FILE`.

I believe that commands have to support this and that built-in support from shells is limited, but I'm not totally sure. That being said, it is very widely supported. In that vein, I can't really find official documentation on it or anything like that: https://unix.stackexchange.com/questions/11376/what-does-double-dash-mean-also-known-as-bare-double-dash

It is mentioned on the `bash` man page as an argument for `bash`.

This is also useful (or required) in cases where an option can take more than one word such as with `--reporters`.

`jest-cli` uses `yargs` internally for its argument parsing. `yargs` doesn't talk about this feature in its main documentation, but it does support it. They talk about it here: https://github.com/yargs/yargs/blob/master/docs/tricks.md#stop-parsing

PR #281, opened 2018-03-01
Add TypeScript support to Karma test file pattern

arusahni:
With the most recent Jest MR (#280) came TypeScript support. This introduces the same for Karma.


janko-m:
Looks good, thanks for the PR!

PR #283, opened 2018-03-07
Fix neoterm integration

kassio:
Neoterm changed [some functions signature](https://github.com/kassio/neoterm/commit/9649cb1de39e398a571b7ea7664843e1fa31fdcb), this will fix the integration.


codeinabox:
Thank you for this, am a big fan of Neoterm too. Will this work with older versions?

Issue #282, opened 2018-03-02
Feature request: Vim8/MacVim terminal support


venables:
One of my favorite vim-test strategies is the neovim terminal.  Vim8/MacVim offers a similar function to Neovim's `termopen`, called `term_start` , which would be fantastic to use.

Based on the [fzf.vim change for this](https://github.com/junegunn/fzf.vim/issues/416), I've hacked a partially-functioning version of this together myself by modifying the neovim strategy:

```
function! test#strategy#vimterm(cmd) abort
  call term_start(a:cmd)
  au BufDelete <buffer> wincmd p " switch back to last window
  startinsert
endfunction
```

This partially works, though the `tern_start` command does not work properly if you assign env-vars as part of your executable, like `NODE_ENV=test mocha`.

janko-m:
@venables Looks great! Would you mind sending a PR with this strategy?

PR #284, opened 2018-03-08
Add Vim8 terminal strategy

venables:
This adds a `vimterminal` strategy which uses the Vim8/MacVim terminal.  

Relates to #282


codeinabox:
Thank you for adding this! Could you please also update `doc/test.txt`

venables:
@codeinabox Good catch -- updated the PR to include doc/test.txt

kassio:
Just a FYI, neovim now supports vim8, so if you want to use vim8 term with vim-test you can use it through neoterm. 😄

codeinabox:
@janko-m any further feedback on this?

PR #286, opened 2018-03-20
cucumberjs: initial implementation

mpareja:
Adds support for running [Cucumber.js](https://github.com/cucumber/cucumber-js) tests from `.feature` files.

I would appreciate a thorough review as this is my first foray into non-trivial vim script authoring.

### Breaking Change

This PR may break existing Ruby users that don't have `.rb` files in their `features` directory. The existing setup had Ruby as the "default" runner for `.feature` files. I considered making the Ruby version search for cucumber in the `Gemfile` but thought that checking for `.rb` files in the features directory would be safer.


autoload/test/ruby/cucumber.vim
```diff
@@ -4,7 +4,7 @@ endif
 
 function! test#ruby#cucumber#test_file(file) abort
   if a:file =~# g:test#ruby#cucumber#file_pattern
-    return empty(glob('features/**/*.php'))
+    return empty(glob('features/**/*.php')) && empty(glob('features/**/*.js'))
```
janko-m: I think we actually *do* want to activate CucumberJS when there are JS files in the `features/` directory, so I think the condition should be negated: `&& !empty(glob('features/**/*.js'))`

I was thinking that, just like Behat (the PHP version of Cucumber) is activated when there are PHP files in the `features/` directory, we could activate the Ruby Cucumber only if there are Ruby files in the `features/` directory, and CucumberJS only if there are JS files. What do you think?

mpareja: I stuck with the approach that was there because I'm not clear how vim-test deals with scenarios where multiple runners qualify. For example, how would vim-test address having _both_ `.rb` and `.php` files under the `features` directory? Does a capability currently exist to help address this conflict?

codeinabox: For JavaScript we can check in package.json, is that sufficient or do we need something similar to check composer.json for PHP or the gem lock file for Ruby?

janko-m: > For example, how would vim-test address having both .rb and .php files under the features directory?

I'm just hoping that no setup will use two languages for implementing Cucumber steps 😃. I think we can ignore this possibility for now, and revisit it if someone opens a ticket.

mpareja: Okay folks, I've updated the algorithm for selecting which cucumber runner to execute.

mpareja: I'm scared of breaking existing Ruby users. The existing setup essentially has Ruby as the "default" runner for `.feature` files. I considered making the Ruby version search for cucumber in the `Gemfile` but thought that checking for `.rb` files in the `features` directory would be safer.

mpareja:
I've updated the PR description to include a Breaking Change section.

PR #288, opened 2018-03-21
Add an option to disable alternate file integration

jeromedalbert:
This PR adds an option to disable alternate file integration.

There are a few reasons why I don't like alternate file integration. One of them is that I use heavily lazy-loaded and customized versions of vim-rails and projectionist to boost vim start-up times. Which makes vim-test bug, and since it is an independent plugin, it should not.


janko-m:
@jeromedalbert Thanks! I just renamed the variable to `test#no_alternate` to match the naming convention of other variables.

PR #300, opened 2018-04-14
Add option to not load default runners

juanibiapina:
Add option to not load default runners

This adds `test#merge_default_runners` with a default value of 1 to keep
the current behavior. It can be set to 0 to disable all default runners.

This is useful if you want to include your own runners but the default
runners end up being picked up instead.

### DIFF
--- a/README.md
+++ b/README.md
@@ -376,6 +376,13 @@ First, add your runner to the list in your `.vimrc`:
 let test#runners = {'MyLanguage': ['MyRunner']}
 ```
 
+Your runner will then be included in the list of runners. To use only your
+runner (excluding all default runners), set this option:
+
+```vim
+let test#merge_default_runners = 0
+```
+
 Second, create `~/.vim/autoload/test/mylanguage/myrunner.vim`, and define the following
 methods:
 
--- a/doc/test.txt
+++ b/doc/test.txt
@@ -419,6 +419,12 @@ You can instruct test.vim to generate absolute file paths:
   let test#filename_modifier = ':~' " ~/Code/my_project/test/models/user_test.rb
 <
 
+You can disable all default runners using:
+
+>
+let test#merge_default_runners = 0
+<
+
 Test.vim relies on you being cd-ed into the project root. However, sometimes
 you may want to execute tests from a different directory than the current
 working directory. You might have a bigger project with many sub-projects,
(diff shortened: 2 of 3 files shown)


juanibiapina:
added some documentation, although I wasn't sure where to put it.

janko-m:
What do you think about adding a `test#enabled_runners` setting instead, where you can select specific runners that you want to be included in matching (all other runners would be disabled). That way people can still select any runners they want to use. Something like this:

```vim
let g:test#enabled_runners = ['ruby#rspec', 'javascript#jasmine'] " only RSpec and Jasmine runners will be matched
let g:test#enabled_runners = ['ruby#my_custom_runner'] " disable all runners except your custom runner
```

This would also partially solve https://github.com/janko-m/vim-test/issues/185.

juanibiapina:
I guess your solution is fine and it solves both problems.

What I don't like is that an assignment is used to add custom runners, but later more stuff gets merged into that variable. It's a little misleading:

```vim
let g:test#runners = {'MyLanguage': ['MyRunner']}
```

Then `test#runners` will contain all runners. I would expect to override it.

On the other hand my solution also doesn't change this, and changing it would be a breaking change anyway.

juanibiapina:
Using `test#enabled_runners` also creates a different nomenclature. It references the function namespaces, while `test#runners` references the capitalized magic names.

So I'm not sure.

janko-m:
> What I don't like is that an assignment is used to add custom runners, but later more stuff gets merged into that variable. It's a little misleading

I agree that it's a bit misleading. But I think it wouldn't be worth it to change that at this point, as it would be a big backwards incompatibility and feature-wise there would be no gain.

> Using test#enabled_runners also creates a different nomenclature. It references the function namespaces, while test#runners references the capitalized magic names.

We could maybe use the same dictionary format as for `test#runners` instead of the one that I proposed. I think that should work equally well, what do you think?

juanibiapina:
Agreed.

juanibiapina:
@janko-m It turns out using the namespace makes the code much simpler. I'm not extremely comfortable with this solution, but it does solve my problem.

juanibiapina:
I should have opened another PR, but I'm kind of sick right now and not thinking completely well :(

janko-m:
Closing in favour of https://github.com/janko-m/vim-test/pull/301.

PR #301, opened 2018-04-16
Add option to select which runners to use

juanibiapina:
If `test#enabled_runners` is set, only runners from that list will be
loaded.

It turns out using the namespace for the runner name makes the code simpler.

As discussed on #300

### DIFF
--- a/README.md
+++ b/README.md
@@ -395,6 +395,19 @@ function! test#mylanguage#myrunner#executable()
 
 See [`autoload/test`](/autoload/test) for examples.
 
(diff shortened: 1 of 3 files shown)


juanibiapina:
I'm not using this plugin anymore, so if this is not going to be merged I would like to close it.

janko-m:
@juanibiapina Sorry for the delay, I got caught up on other things.

The pull request looks good to me 👍. And I really appreciate it, this will be very useful ❤️

PR #304, opened 2018-05-19
Add g:test#vimterminal_persist_session option

alexgenco:
This option opens an interactive shell in a terminal buffer and sends
test commands using term_sendkeys(). This is useful when the user wants
to have their history of test runs available in a single terminal
buffer, rather than opening a fresh one every time.

This is my preferred behavior, so I figured I would see if others found it useful. Otherwise I'll just implement it as a custom strategy.

### DIFF
--- a/autoload/test/strategy.vim
+++ b/autoload/test/strategy.vim
@@ -56,8 +56,22 @@ function! test#strategy#neovim(cmd) abort
 endfunction
 
 function! test#strategy#vimterminal(cmd) abort
-  botright new
-  call term_start(['/bin/sh', '-c', a:cmd], {'curwin':1})
+  if exists("g:test#vimterminal_persist_session") && g:test#vimterminal_persist_session
+    if !exists("g:test#vimterminal_shell")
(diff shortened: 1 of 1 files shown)


codeinabox:
Thanks for this and LGTM, could you please also update the README and doc/test.txt so others know of this option

@janko-m any further thoughts?

janko-m:
Thanks for the pull request!

However, I would prefer to avoid adding configuration options for specific strategies, I can imagine this growing. Is there some existing plugin for managing Vim's terminal we can utilize (that would have its own configuration)? Or maybe it would be worthwhile creating a new plugin and using that for the vimterminal strategy?

alexgenco:
> However, I would prefer to avoid adding configuration options for specific strategies

That's probably a good idea

> Is there some existing plugin for managing Vim's terminal we can utilize

Not sure about this, but I couldn't find anything from a quick search. Seems like it would basically be vimux but using vim terminal instead of tmux, which seems very useful to me.

I'll use my custom strategy for now, and keep an eye out for a suitable plugin. Thanks!

PR #307, opened 2018-07-10
Small improvements to swiftpm testing

gfontenot:
- Syntax for limiting the cases to run has changed
 - Improve the quality of the test code a bit
 - Improve module recognition

### DIFF
--- a/spec/swiftpm_spec.vim
+++ b/spec/swiftpm_spec.vim
@@ -14,18 +14,18 @@ describe "SwiftPM"
   it "runs nearest tests"
     view +6 Tests/VimTestTests/VimTestTests.swift
     TestNearest
-    Expect g:test#last_command == 'swift test --specifier VimTestTests.VimTestTests/testExample'
+    Expect g:test#last_command == 'swift test --filter VimTestTests.VimTestTests/testExample'
 
     view +10 Tests/VimTestTests/VimTestTests.swift
     TestNearest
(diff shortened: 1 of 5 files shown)


spec/swiftpm_spec.vim
```diff
@@ -1,38 +1,52 @@
 source spec/support/helpers.vim
 
 describe "SwiftPM"
-
-  before
-    cd spec/fixtures/swiftpm/
-  end
-
```
janko-m: Any reason for moving this `cd` statement out of the `before` block?

gfontenot: Explicitness more than anything. I do think that eventually for swift-test we might want multiple fixtures, but I suppose it would still be useful to cd into the root of the fixture folder in the `before` block.

tbh I wrote this like 3 weeks ago before going on vacation and just stumbled across it this morning, so I've mostly forgotten my reasonings. I don't have a super strong opinion on this and am totally happy to drop this if you feel strongly about it.

janko-m: I would prefer to keep it in the `before` block for now, as we're only using a single directory so far. Later we can use `context` to set up different `before` blocks with different folders as needed.

gfontenot: 👍 Done!

janko-m: Great, thanks!

PR #309, opened 2018-07-25
Add vertical neovim strategy

smarquez1:
Add Vertical Neovim strategy to run tests in a terminal in a vertical, instead of the default horizontal.
 
Sorry for this lousy PR, but before I improve it I'd like to know if you are accepting this kind of strategies at the repository.

Other options would be to just document this kind of strategy for the user to add it to their init.vim or to add options to the basic `neovim` strategy. Please, let me know what to you think and I will improve it.

### DIFF
--- a/autoload/test/strategy.vim
+++ b/autoload/test/strategy.vim
@@ -55,6 +55,13 @@ function! test#strategy#neovim(cmd) abort
   startinsert
 endfunction
 
+function! test#strategy#neovim_vertical(cmd) abort
+  vertical new
+  call termopen(a:cmd)
+  au BufDelete <buffer> wincmd p " switch back to last window
+  startinsert
+endfunction
+
 function! test#strategy#vimterminal(cmd) abort
   botright new
   call term_start(['/bin/sh', '-c', a:cmd], {'curwin':1})


janko-m:
Thanks for the pull request, the changes look good to me :+1:

Before I merge, have you checked whether the “neoterm” plugin already has such a feature? Vim-test has a strategy for neoterm, and since neoterm is highly configurable I’d prefer users leveraging it over adding new strategies to vim-test.

codeinabox:
According to the Neoterm documentation

> Revamp :Topen. Now :[mods][N]Topen accepts vim mods (:h mods) and a target, so if one wants to open the neoterm with id 2 in vertical, one can do :vert 2Topen.

janko-m:
Thanks for looking into it! Ok, I was hoping that neoterm has the feature to turn it on by default, not that it would require a separate command.

In that case this PR makes sense to me. Would you mind updating the README and doc/test.txt?

zhaochy1990:
is there any update on this?

smarquez1:
I'm sorry everyone, I totally forgot about this. I'll improve the PR next week so we can merge it.

janko-m:
This PR might be outdated, since https://github.com/janko-m/vim-test/pull/319 allows the `neovim` terminal position to be configurable

smarquez1:
Closing this ticket since #319 allows to configure a vertical terminal using the `neovim` strategy. Here's an example of how you can achieve this:

```
let test#strategy = "neovim"
let test#neovim#term_position = "vert"
```

PR #310, opened 2018-07-26
fix(csharp): add missing flag arg in substitute cmd

jmzagorski:
This PR adds the missing flags argument in the `substitute` command that only seems to be problematic in windows. The reason I probably missed this is all tests were run on WSL on windows 10 and passed. I was even able to run actual commands in a dotnet project. However, moving over to windows I get the error below that is fixed with this PR, I really wanted to create a test against the csharp.vim file that I could run on windows, but I could not get vim-flavor to run on windows.

image


janko-m:

PR #312, opened 2018-08-15
Add minimal scala support

axrt:
Hey janko-m,

First of all, thank you for the brilliant work, I am a huge fan, and use vim-test regularly to test my python, ruby and java!
Should you choose to include my contribution for scala testing through sbt, it would be my greatest pleasure.
I tested the solution, and it works fine for me, but as scala imposes way less formatting restrictions compared to java tests for instance, it is hard to create a file pattern regex to match all possible cases. If (when) I find a better one, I will definitely let you know.
Also, I am mostly using maven as scala is a mixin, rather than my primary tool. I will cut some time in the future to attach maven test to scala as well.

Thank you once again, 

with my best wishes, 

Alex T

### DIFF
--- a/autoload/test/scala/sbttest.vim
+++ b/autoload/test/scala/sbttest.vim
@@ -1,10 +1,11 @@
 if !exists('g:test#scala#sbttest#file_pattern')
-  let g:test#scala#sbttest#file_pattern = '\v^.*.scala$'
+  let g:test#scala#sbttest#file_pattern = '\v^(.*test.*|.*suite.*)\c\.scala$'
 endif
 
 " Returns true if the given file belongs to your test runner
 function! test#scala#sbttest#test_file(file) abort
-  return a:file =~? g:test#scala#sbttest#file_pattern
+  let current_file = fnamemodify(a:file, ':t')
+  return current_file =~? g:test#scala#sbttest#file_pattern
 endfunction
 
 " Returns test runner's arguments which will run the current file and/or line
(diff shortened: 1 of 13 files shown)


plugin/test.vim
```diff
@@ -24,6 +24,7 @@ let g:test#default_runners = {
   \ 'Perl':       ['Prove'],
   \ 'Racket':     ['RackUnit'],
   \ 'Java':       ['MavenTest'],
+  \ 'Scala':      ['Sbttest'],
```
janko-m: I think it would make more sense to have this named `SbtTest`.

autoload/test/scala/sbttest.vim
```diff
@@ -0,0 +1,39 @@
+if !exists('g:test#scala#sbttest#file_pattern')
+  let g:test#scala#sbttest#file_pattern = '\v^.*.scala$'
```
janko-m: Is it possible to make the test file regex match test files but not application files, e.g. `\vsrc/test/scala/.*\.scala$`? That helps make test.vim smarter, e.g. if you run tests on an application file, it would detect that it's not a test file and run tests for the last test file instead.

plugin/test.vim
axrt: Fixed)

autoload/test/scala/sbttest.vim
axrt: Good point, but I looked at our codebase at work, and it looks like it is overwhelmingly common for people to include "Test", "Suite", or both in the filename. I will work on an improved regex; so far, I created a fixture for scala, will create a spec and extend README tomorrow.

axrt:
Ok, Travis is finally happy:)
I still did not figure out what to do about the messy files, like if the file is named differently and/or has more than one class defined (which scala inconveniently allowed), but it looks like the best way to go would be to pass down the class name as a parameter.

I guess it's good for now.
Thanks for reviewing!

janko-m:
Thanks! Great job with the tests. Merging :shipit:

PR #339, opened 2019-01-02
Add the ability to configure Vim's terminal position

sstutz:
This PR adds a way to configure the terminal position for vim's built-in terminal.
It'd also add the new test#vim#term_position variable which IMHO is redundant since the same-ish variable exists for neovim. 

Personally I think it would be better to have one test#vim#term_{split,position} variable and use it for both, vim and neovim.

but this decision is up to @janko-m

### DIFF
--- a/README.md
+++ b/README.md
@@ -253,15 +253,16 @@ let test#ruby#rspec#options = {
 \}
 ```
 
-### Neovim terminal position
+### Vim8 / Neovim terminal position
 
-The `neovim` strategy will open a split window on the bottom by default, but
+Both the `neovim` and `Vim8 Terminal` strategy will open a split window on the bottom by default, but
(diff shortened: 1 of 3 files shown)


sstutz:
@janko-m I was asking you whether you want me to refactor it to have a shared variable between vim/nvim.
In case that wasn't obvious :)

janko-m:
@sstutz I agree about the shared variable. For now let's merge this PR, and we can add the common variable subsequently. At the moment I don't see the benefit of splitting it into two variables (split, position), but let me sit on it.

In the meanwhile, could you rebase to `master`, and possibly update the README?

sstutz:
done

janko-m:
Perfect, thank you!

PR #341, opened 2019-01-16
vimterminal: switch back to previous window upon exiting the test runner buffer

jonleighton:
I initially tried to use BufDelete as per the neovim strategy, but it
didn't work. BufLeave works well for me.


jonleighton:
Rebased onto https://github.com/janko-m/vim-test/pull/342, so you can merge that one first and then this one :+1:

jonleighton:
Or just merge this one... in hindsight it probably wasn't the best idea to create three separate PRs for this stuff!

janko-m:
Thanks! I think it was nice that you separated the changes, so that we can discuss them and merge them individually. They just all happened to modify the same part of the code 😃

PR #342, opened 2019-01-16
vimterminal: quit on <Enter>

jonleighton:
This make it behave more like the basic strategy.

I'm putting this up for discussion - it might be a bad idea to implement as a default, what do you think? I have found it a helpful shortcut.


janko-m:
I like it, thanks for the patch!

I was hoping that the insert mode in Vim terminal will have the same functionality as in Neovim (pressing any key will close the terminal), but it does appear `:q` is the way to close the terminal.

I think this functionality is very convenient, as `<Enter>` isn't used for anything useful as far as I know. So I'm 👍for merging. Would you mind rebasing?

jonleighton:
Done :+1:

PR #347, opened 2019-01-30
Vimterminal enhancements

jonleighton:
This is a series of related enhancements to the vimterminal strategy. I've been using this setup for about a week and I think it's working really nicely. I have `:TestClose` mapped to `<leader>e`, and usually I don't actually enter the terminal window at all - I just fire tests off and just hit `<leader>e` when I want the window to go away.

---

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,7 @@ runners are supported:
 | **Elm**        | elm-test                                                                                    | `elmtest`                                                                                                       |
 | **Erlang**     | CommonTest, EUnit                                                                           | `commontest`, `eunit`                                                                                           |
 | **Go**         | Ginkgo, Go, Rich-Go, Delve                                                                  | `ginkgo`, `gotest`, `richgo`, `delve`                                                                           |
+| **Haskell**    | stack                                                                                       | `stacktest`                                                                                                     | 
 | **Java**       | Maven, Gradle                                                                               | `maventest`, `gradletest`                                                                                       |
 | **JavaScript** | Ava, Cucumber.js, Intern, Jasmine, Jest, ReactScripts, Karma, Lab, Mocha, TAP, WebdriverIO  | `ava`, `cucumberjs`, `intern`, `jasmine`, `jest`, `reactscripts`, `karma`, `lab`, `mocha`, `tap`, `webdriverio` |
 | **Lua**        | Busted                                                                                      | `busted`                                                                                                        |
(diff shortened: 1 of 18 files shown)


janko:
Thank you for the improvements! I will review the changes this week.

codeinabox:
@janko would you like a hand reviewing this? @jonleighton if you could resolve the conflicts that would be great, thank you for your patience

janko:
@codeinabox Yes, please 👍

jonleighton:
@codeinabox Done.

As a side note, I've recently started using Neovim. Would you be open to a PR which brings the neovim strategy into line with this?

The main thing I want to change with the current neovim strategy is that it closes the window when any key is pressed, once the test run has finished. This means that if there's any output which requires scrolling to see (e.g. a backtrace from a failure), then I can't actually see that information. So I'd propose implementing a `:TestClose` command, like in here, and then the user has the option to decide how they want to close the test output.

codeinabox:
@jonleighton yes that would be great as I use Neovim myself

janko:
@codeinabox Let's first properly review and merge this change before making it for Neovim. I want to make sure these changes are not introducing undesirable side effects (i.e. something that was convenient with default behaviour but now it's not). That's the reason why I was holding off on merging this PR, because I wasn't sure whether the new strategy would remain equally convenient to users that might not have need for the features this PR brings.

We should discuss behaviour before & after. For example, I think originally the terminal window would be focused, and from what I see with these changes it's not anymore. What are the pros and cons for that? If people want to scroll through the test output, how easy it is for them to switch to the terminal window? Maybe with 2x2 grid of windows it's easy, is it the same for a 3x3 grid? We should evaluate things like that.

emilford:
:wave: Hi, @janko.

Are these changes still up for consideration?

I'd similarly prefer the terminal not to gain focus, as well as a way to close the terminal from the outside. Understanding that the current behavior may be preferred by some, could this be configurable option instead? I would ideally continue to use the `vimterminal` strategy, but have recently reverted to `vtr` because of some of what this PR would solve for.

janko:
@emilford Historically I have preferred to avoid adding configuration variables for strategies, and delegated that to the plugin that activates the strategy. In this case, `vimterminal` uses Vim terminal directly, but if you want additional configuration, then `vtr` is preferred, because it likely has this configuration.

jonleighton:
Almost three years later I have come back to this :upside_down_face: 

Given that this is not going to be merged, I have extracted the functionality to a separate plugin that implements it as a custom strategy: https://github.com/jonleighton/vim-test-vimterminal-enhanced

So if you want to use that @emilford you could just drop in the plugin rather than using my fork.

I'll close this PR now.

Issue #74, opened 2015-11-09
TestNearest won't work with @test annotation


mjacobus:
TestNearest wont work with test annotation:

Vim test won't recognize that functions annotated with "@test" are also test. Therefore when `TestNearest` is run it will behave like `TestFile`.
```php
/**
 * @test
 */
public function foo()
{
    $this->assertTrue(false); # cursor here
}
```

The below code works

```php
public function testFoo()
{
    $this->assertTrue(false);  # cursor here
}
```

If you accept I can try to write a PR. Any tips on how to run the tests?

janko-m:
Hey, thanks for reporting. Yes, I would accept a PR for this. If you have Ruby installed, it's just a `gem install vim-flavor` and you can run `:VSpec` when inside of Vim. Alternatively I could add Travis CI which would run the tests for you.

The only problem is that `test#base#nearest_test()` can currently only find matches on the line where the test name is seen. So, if it sees `@test`, you cannot tell it to go down until the function definition. Maybe the function could be extended somehow, alternatively you could copy that behaviour to the Maven runner and alter it to do what you want.

mjacobus:
Thank you @janko-m! I will give that a try =)

PR #349, opened 2019-02-03
Fix phpunit nearest

elythyr:
Hi again,

I don't think you will remember this almost three years old issue https://github.com/janko-m/vim-test/issues/74
But the fix that was proposed was actually a hack that look for the first public method instead of the actual annotation.
It works great most of the time, I mean like 99% of the time!

Unfortunately since PHP7 we are allowed to create what is called anonymous classes.
Short story long, it's a way to create a class wherever we might want, even inside a method.

I faced such a case this afternoon and realized that if my cursor is inside my test method but under the anonymous class the `:TestNearest` command will stop on the public method of the anonymous class and try to gave it as a filter to PHPUnit.

So I propose this PR to correct that unfortunate behavior.
This is a two steps job and the first one might seems dangerous, all depends on how much you trust the tests :)

I have refactor the `test#base#nearest_test` function to use a new function named `test#base#nearest_test_in_lines`.
This function does the exact same thing but is not coupled with the `position` argument and therefore can also look forward.
Plus, in the returned dictionary I added a new key `test_line` to be able to get the line where the test pattern was matched.

All tests passes, so it should not have break anything.
Feel free to ask for more information, I didn't wan to put too much things inside this message so it stays relatively short and you can find some more explanation inside the code as comments and in the commits messages.

P.S.: Once again, no updates of the README nor the doc because there is no new functionality and it should not even be consider as a change in the behavior.

### DIFF
--- a/autoload/test/base.vim
+++ b/autoload/test/base.vim
@@ -97,7 +97,7 @@ function! test#base#nearest_test_in_lines(filename, from_line, to_line, patterns
   let current_line = a:from_line + 1
   let test_line    = -1
 
-  let is_reverse = '$' == a:from_line ? v:true : a:from_line > a:to_line
+  let is_reverse = '$' == a:from_line ? 1 : a:from_line > a:to_line
   let lines = is_reverse
     \ ? reverse(getbufline(a:filename, a:to_line, a:from_line))
     \ : getbufline(a:filename, a:from_line, a:to_line)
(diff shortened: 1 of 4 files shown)


janko:
Thanks a lot for the pull request! I haven't forgotten about it, the other day I started reviewing it, and what I read seemed good. I will try to finish my review soon and merge the pull request.

elythyr:
Just checking in to see if it's still up to date to merge it ?

PR #350, opened 2019-02-04
Remove superfluous help markup

sodapopcan:
This just removes an extra `<` in the help file.


janko:
Thanks for the patch!

PR #351, opened 2019-02-04
Fix markup in help file

sodapopcan:
Found another one.  I can squash this with the last one if you like.


sodapopcan:
Oh, you've already merged the other!  Sorry, I shouldn't have been so hasty to send the previous PR before finishing my doc-reading session.

janko:
Np, thanks for another patch!

Issue #361, opened 2019-04-12
TestNearest failed with annotated PHP test


jesseleite:
I can still test nearest on this...
```php
public function test_it_still_does_a_thing()
{
    // this still works
}
```

But can no longer test nearest on this...
```php
/** @test */
public function it_does_a_thing()
{
    // no longer works
}
```

The latter is a valid PHPUnit convention, and worked until I recently updated vim-test 🤔 

Thoughts on this?

janko:
That could be related to https://github.com/janko/vim-test/pull/349. @elythyr any ideas?

elythyr:
Hi there,

It seems this way, after removing the merge from #349 there is no error anymore.

This seems to be because your phpdoc is on only one line.
If you pass it on multple lines:
```php
/**
 * @test
*/
```
It will work fine.
Just the time to fix it, will add a new test for it.

Will try to propose you a fix in the afternoon.

jesseleite:
Ah cool, thanks for looking into!

janko:
This should be fixed by https://github.com/janko/vim-test/pull/362

PR #362, opened 2019-04-13
[Bugfix - #361] Handles "/** @test */" on one line

elythyr:
The fix for the ticket #361 
It was only missing on test for this case.

For phpunit the conditions to find a test was a public method with a name starting by "test".
Or the public method right after a phpdoc block containing the tag "@test", in the form:
```php
/**
* @test
*/
```

I just added a new test to take the following case into account:
```php
/** @test */
```

I'm not use to this style so I don't know if it's tolerated to have multiple tags one the same line...
For now it only works for this specific case.
The reason behind that is that I don't think it's a good idea to have multiple tags on the same line, I'm not event sure it's tolerated in phpdoc.
Plus, we got a great tool called Vim which provides folding :D

I added a new test to confirm the error before fixing it.
And I realized I had commented the ones from the previous commit...
So I put them back on.

### DIFF
--- a/autoload/test/php/phpunit.vim
+++ b/autoload/test/php/phpunit.vim
@@ -3,10 +3,15 @@ if !exists('g:test#php#phpunit#file_pattern')
 endif
 
 if !exists('g:test#php#phpunit#test_patterns')
+  " Description for the tests:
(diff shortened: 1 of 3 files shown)


jesseleite:
Thank you both again!

PR #369, opened 2019-06-07
Extends pytest to support nested test classes

willjp:
- [x] `(n/a)` Update the README accordingly
- [x] `(n/a)` Update the Vim documentation in `doc/test.txt`

Hello again!

I added a couple more changes, this time to pytest if they are useful for you. Give me a shout if you don't like anything, I'm happy to make changes if this doesn't feel right. Have a great weekend!

* `:TestNearest` now also works on pytest nested test classes
* Most previous tests were using pytest to run xunit tests. I added a few test cases for pytest as both a test runner, and a test framework.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,12 +2,12 @@
 
 A Vim wrapper for running tests on different granularities.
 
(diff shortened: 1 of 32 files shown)


janko:
I apologize, this has slipped my radar. The changes look good, and the test failure is unrelated, thank you! ❤️

PR #378, opened 2019-07-06
Feature/kitty strategy

metalelf0:
Hi @janko , I'm working on a [kitty terminal](https://github.com/kovidgoyal/kitty) strategy for vim-test. This first working version uses the same approach of the iterm2 strategy; it also requires the user to set an environment variable and launch kitty with a listen socket. Let me know what you think.

Keep up the great work!


janko:
Looks good to me, thank you for the contribution! 👍

Before I merge, would you mind making "Kitty" in the README a link to the website/repo (just so that people can easily see what it's about if they don't know), and updating `doc/test.txt` as well?

metalelf0:
Ok, sure! Should be ok right now. :)

PR #384, opened 2019-08-14
add poetry support to pytest

fishman:


autoload/test/python/pytest.vim
```diff
@@ -44,6 +44,10 @@ function! test#python#pytest#executable() abort
     let pipenv_prefix = "pipenv run "
   endif
 
+  if filereadable("poetry.lock")
```
janko: Could this be an `elseif` chained on the previous conditional? That would be more intention revealing, because either Pipenv or Poetry will be used, never both.

janko:
@fishman Thank you, merging away 🚀

PR #385, opened 2019-08-16
Add support for the Delve test runner for Go

leighmcculloch:
# What
Add support for the Delve test runner for Go that will run `dlv test`
for the identified tests and include breakpoints and tracepoints if
`vim-delve` is also in use.

# Why
`vim-test` makes it really easy to run tests in Go programs, but it's
much harder to debug tests. In other languages like Ruby it's relatively
easy because their debuggers are inserted into their normal application
runtimes, but with Go to debug requires executing a different
application with unique parameters. This makes debugging Go tests as
simple as running them, all with vim-test.

Additionally, if vim-delve is in use a user can select breakpoints and
tracepoints within vim, and this change includes pulling in that list of
instructions and includes them in the init setup of dlv. Use of this component
is optional and you can quite reliably use dlv test without passing
that init through by defining breakpoints at the dlv console.

# Note
Part of this change is dependent on sebdah/vim-delve#34.

One change was made to options to allow a test runner to modify the
options before including them into the cmd to be executed. The delve
command is a little odd in that it specifies the package part of the
build position before all other flags.

# Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ runners are supported:
 | **Elixir**     | ESpec, ExUnit                                                                               | `espec`, `exunit`                                                                                               |
 | **Elm**        | elm-test                                                                                    | `elmtest`                                                                                                       |
 | **Erlang**     | CommonTest                                                                                  | `commontest`                                                                                                    |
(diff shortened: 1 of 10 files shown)


janko:
Thank you for the pull request! ❤️

The vim-test change regarding ordering options looks very clean and safe, thanks for that. Merging away 🚀

PR #386, opened 2019-08-16
Fix options being appended in delve runner

leighmcculloch:
# What
Insert a `--` in the dlv test command between the package name and where
test options are supplied.

# Why
The `--` is necessary to tell the dlv test command where options for go
test can be supplied vs where the package list start and ends.

When I originally developed the delve runner I had included a `--` and
then removed it thinking it was unnecessary just before I opened the PR
(#385). The test scenarios I'd coded didn't pickup on the options now
being ignored.

Sorry @janko, I didn't pickup on it fast enough before it was merged!

# Checklist

- [ ] ~Update the README accordingly~
- [ ] ~Update the Vim documentation in `doc/test.txt`~

### DIFF
--- a/spec/delve_spec.vim
+++ b/spec/delve_spec.vim
@@ -16,14 +16,14 @@ describe "Delve"
     view +5 normal_test.go
     TestNearest
 
-    Expect g:test#last_command == 'dlv test ./. -test.run ''TestNumbers$'''
+    Expect g:test#last_command == 'dlv test ./. -- -test.run ''TestNumbers$'''
   end
 
   it "runs nearest tests in subdirectory"
     view +5 mypackage/normal_test.go
     TestNearest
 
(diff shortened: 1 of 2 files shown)


spec/delve_spec.vim
```diff
@@ -61,11 +61,17 @@ describe "Delve"
...
     end
-    it "appends options to the end"
+    it "appends options to the end in a suite test"
       view normal_test.go
       TestSuite
 
-      Expect g:test#last_command == 'dlv test ./... -test.v'
+      Expect g:test#last_command == 'dlv test ./... -- -test.v'
+    end
+    it "appends options to the end in a nearest test"
```
leighmcculloch: This test exists to ensure the `--` is not added twice, since it can be added between the package name and the `-test.run` which are power of the bills position, as well as before the custom options `-test.v`.

PR #391, opened 2019-09-01
Recognize `Task` as well as `void` return type

tmadsen:
It's not uncommon to have tests have a `Task` return type when they're
async, instead of void. With this patch, `:TestNearest` works as
intended when the cursor is inside a test like this:

```
[Fact]
public async Task MyTest()
{
    Assert.True(true)
}
```

Before this patch, `:TestNearest` would test the whole file instead.


janko:
Looks good, thanks!

PR #394, opened 2019-09-23
Support `spec` files as well with sbt

yannick-cw:
I could not figure out at all, why my scala test files were always run with `bloop`. I think sbt should also be allowed to run fields with `Spec` in the name


janko:
Looks good, thank you 👍

PR #407, opened 2019-10-30
Add support for testrunner configuration via vim-projectionist

### LINKED ISSUE #406: Make testrunner configurable via vim-projectionist
While working on multiple java project which either used gradle or maven I noticed, that it would be awesome to configure the used testrunner via vim-projectionist. 

After thinking about it for a while I came up with something like this:

```json
{
   "*": {
      "test-runner": {
          "java": "gradle"
       }
   } 
}
```

Are there any reasons why this should not be done?

patrickpichler:
It is now possible to configure testrunners via a vim-projectionis
configuration file. This allows for an easy per project basis
configuration of your testrunners.

This fixes #406

### DIFF
--- a/README.md
+++ b/README.md
@@ -498,6 +498,42 @@ You can disable this integration by doing
 let g:test#no_alternate = 1
 ```
 
+It is also possible to configure test runners via projectionist. All you 
+need is to add a `test-runner` key to the projectionist configuration. 
+The value for `test-runner` has to be a dictionary containing the target
+languages as keys and either the name of the runner or an dictionary 
+with additional configuration as value.
+
+Here an example:
+
+```json
+{
+  "gradle/*Test.java": {
+    "test-runner": {
+      "java": {
+        "runner": "gradletest",
+        "options": {
+          'nearest': '--backtrace' 
+        }
+      }
+    }
+  },
+  "mvn/*VerboseTest.java": {
+    "test-runner": {
+      "java": {
+        "runner": "maventest",
+        "options": "--verbose" 
+       }
+    }
+  },
+  "mvn/*Test.java": {
+    "test-runner": {
+      "java": "maventest" 
+    }
+  }
+}
+```
+
 ## Extending
 
 If you wish to extend this plugin with your own test runners, first of all,
--- a/autoload/test.vim
+++ b/autoload/test.vim
@@ -3,16 +3,15 @@ function! test#run(type, arguments) abort
 
   let alternate_file = s:alternate_file()
 
-  if test#test_file(expand('%'))
-    let position = s:get_position(expand('%'))
-    let g:test#last_position = position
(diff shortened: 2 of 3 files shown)


janko:
Thanks for submitting the pull request! I will try to find time soon to look at it in more detail.

lourenci:
This patch is not working for me. Here is my configuration:
```
{
  "*.spec.js": {
    "alternate": "{}.js",
    "test-runner": {
      "javascript": "jest"
    }
  }
}
```
The vim-test heuristic tries to run the karma test runner.

patrickpichler:
I have tested it mainly with gradle/maven as this was my usecase. With those you can set the `g:test#java#runners` variable to configure which runner to use. I just noticed that the js test runners try to detect which test framework you have installed via the projects `package.json`. 

This was clearly an lack of testing on my side. 

The question is how we should resolve this. Either we implement the check for the `g:test#javascript#runner` variable in the js test runners (and all the others) as well, or come up with some global mechanism to do this dynamically. And should it be possible to override the autodetection via the `package.json`?

codeinabox:
> The question is how we should resolve this. Either we implement the check for the `g:test#javascript#runner` variable in the js test runners (and all the others) as well, or come up with some global mechanism to do this dynamically. And should it be possible to override the autodetection via the `package.json`?

I think what you want is some logic that works like that this pseudo code:

> if test-runner is defined in projections OR package.json contains blah then use this runner

Let me know if you need any help with this as I worked on the package.json logic

patrickpichler:
I would like to generalize the logic. It would be nice if each runner would define a method returning language it is for and a second one the name it has. Then we can filter all runners according to the projectionist values + return values of the new methods. The resulting runners can then be processed as before. 

This way we would not break existing configurations (I hope i didn't miss something).

patrickpichler:
This is still a WIP but he new version now filters the runners before testing them. This should fix the issue with the nodejs runnres. 

Another change is that the projectionis settings won't set global variables. This should prevent some strange side effects when using the projectionist settings (leaking of settings). 

There is still some work to be done. 
 - The code should probably be refactored a bit
 - Look through the code of all test runners and see if this change breaks anything

patrickpichler:
@lourenci could you retest if your usecase works?

lourenci:
@patrickpichler, for the test case I've posted it is ok. :) I'll use your branch in my daily routine and keep you posted on something wrong. Thank you.

patrickpichler:
After looking through the runners, I notice that quite a few are using
some global variables. This could lead to some strange behaviours. The
big issue is, that fixing most of this usages would break existing
workflows. 

I think it would be best to add basic support for vim-projectionist with
this PR and introduce the other changes to the test runners in their own
PRs.

patrickpichler:
@janko could you maybe look over the code and provide some feedback?

PR #410, opened 2019-11-06
support reading package.json from non-root cwd

mbriggs:
This is an extremely minor change to allow for reading package.json in a non root location. I am doing client work on a massive repository, and frequently will `:cd` into a project directory when I am working on it in order to have a sane fzf experience :)

### DIFF
--- a/autoload/test/javascript.vim
+++ b/autoload/test/javascript.vim
@@ -4,7 +4,8 @@ let test#javascript#patterns = {
 \}
 
 function! test#javascript#has_package(package) abort
-  for line in readfile('package.json')
+  let json_path = findfile('package.json', '.;')
+  for line in readfile(json_path)
     if line =~ '"'.a:package.'"'
       return 1
     endif


autoload/test/javascript.vim
```diff
@@ -4,7 +4,8 @@ let test#javascript#patterns = {
 \}
 
 function! test#javascript#has_package(package) abort
-  for line in readfile('package.json')
+  let json_path = findfile('package.json', '.;')
```
codeinabox: We should get this to exclude node_modules as that will contain main package.json files

codeinabox:
Thank you for this pull request. My only reservation is we might need to update all the runners as they presume that the node_modules directory which can contain the test runner binary is in the current working directory.

ShahinSorkh:
I guess you can use `npx` or `yarn dlx` to run the runner instead of checking a hard-coded path.
It will check the project root if find one, and then will install the requested runner if is not installed as a dependency.

ShahinSorkh:
`yarn dlx`: https://yarnpkg.com/cli/dlx
`npx`: https://www.npmjs.com/package/npx

_Note:_ `npx` is usually bundled with `npm`.

fnune:
This patch makes `vim-test` work on my monorepo!

adrianhelvikspond:
For jest it looks like it doesn't call jest from the correct directory. So while it finds `package.json` and determines the test runner correctly, I get the error `command not found: jest`.

PR #412, opened 2019-11-17
Remove --no-coverage flag from jest

deriegle:
A `--no-coverage` flag was being added to every test command for Jest,
but I don't see this option in their configuration and it was causing
issues with one of my repos.

Is there a reason this was added? I'd be interested in trying to add an option that conditionally adds this if it's necessary.

https://jestjs.io/docs/en/cli

_I removed 2 of the checkboxes, because they did not apply to this change. Let me know if I should implement doc changes for this._

### DIFF
--- a/autoload/test/javascript/jest.vim
+++ b/autoload/test/javascript/jest.vim
@@ -13,9 +13,9 @@ function! test#javascript#jest#build_position(type, position) abort
     if !empty(name)
       let name = '-t '.shellescape(name, 1)
     endif
-    return ['--no-coverage', name, '--', a:position['file']]
+    return [name, '--', a:position['file']]
   elseif a:type ==# 'file'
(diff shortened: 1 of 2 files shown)


janko:
This flag was added in https://github.com/janko/vim-test/pull/291 by @mpareja. Does this option not exist in some jest versions?

mpareja:
There is a [---coverage](https://jestjs.io/docs/en/cli#coverage) option. The `no` prefix disables the configuration setting. Can you elaborate on the "issue" it was causing in _one of your repos_?

deriegle:
Thanks everyone for looking at this so quickly! I realized after putting this PR up that we really needed to fix a script we're using to run tests. Our VSCode plugin was putting the file path first, followed by any flags and this plugin is putting the file path as the last argument. I just updated our script to handle either case.

PR #424, opened 2020-02-01
Add ParaTest (PHP) support

partounian:


codeinabox:
Thank for you this as I hadn't know about ParaTest prior to this. I noticed the tests are failing because currently there is no way to distinguish whether to use the `paratest` command or `phpunit` - any thoughts on how to do this? From what I can tell, one way is to check if the binary is present. Also is there any benefit for using ParaTest for test nearest and test file or are the benefits only when running the whole test suite?

partounian:
The default seems to be running per class, while adding the `-f` flag will make it parallel per method. Though they state, that this makes it slower as then you need to bootstrap each method rather than each class. I can't see how there would be a benefit to using ParaTest for a single method.

I just made adjustments but tests are still failing, though for a different reason. My cheap way of adjusting the test is not that great either.

partounian:
P.S. there is also https://github.com/facile-it/paraunit and I just started using paratest so no idea if there's a difference.

codeinabox:
Do ParaUnit and ParaTest take the same command line arguments as PHPUnit? If so I might create a PR that changes the behaviour of the PHPUnit runner to use ParaUnit or ParaTest if present

partounian:
I can't say with confidence as I am new to both tools.

partounian:
Seems like (Laravel's) `dd($var)` && `var_dump($var)` don't work in paratest. I still would like to implement this and also make it easy to turn the "para"s off in favor of phpunit or by default always run phpunit for testnearest.

Do you have any suggestions on how I could properly implement this?

codeinabox:
Happy to help you with this but first I need to understand what you are trying to achieve here and the uses cases. This my current understanding so please correct if I am wrong:

- PHPUnit does not run tests in parallel but there are several alternative parallel test runners including ParaTest and Paraunit
- It appears that running tests in parallel is most beneficial for test suites
- ParaTest states that running some test classes in parallel involves using functional mode that may actually be slower https://github.com/paratestphp/paratest#optimizing-speed
- You have also stated that some commands don't work with ParaTest and you like to make it easy to configure when it is used

Based on that understanding it is my suggestion that we allow users to override the binary used for running PHPUnit test suite. This gives the user the flexibility you want and avoids the need to have add a runner for every alternative PHPUnit runner. What are your thoughts on this?

@janko I was thinking of adding a config option like `test#php#phpunit#executable_suite` perhaps

partounian:
@codeinabox 

Your 4 points are correct.`dd` is basically Laravel's pretty var_dump and exit the suite, while var_dump is a php stdlib var dump.

That config option doesn't seem bad though it seems that it usually best to use a parallel test runner except for TestNearest, in that case simply running `phpunit` is preferred for the reasons of being able to use `dd` and no unnecessary overhead.

If we could set this as the default if the user has a parallel testing suite, I believe that would be best.

codeinabox:
I am going to see if I can add paratest / paraunit support when I create a feature for #445 as it makes sense to have alternative binaries for the phpunit runner as long as they accept the same arguments

codeinabox:
Thank you for your patience on this. I have a created a pull request with an alternate solution if you could give that a try.

PR #425, opened 2020-02-01
Fix Ruby Minitest command

nomasprime:

### DIFF
--- a/autoload/test/ruby/minitest.vim
+++ b/autoload/test/ruby/minitest.vim
@@ -76,9 +76,9 @@ function! test#ruby#minitest#executable() abort
     endif
   else
     if filereadable('Gemfile') && get(g:, 'test#ruby#bundle_exec', 1)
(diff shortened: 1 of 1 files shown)


janko:
Which issue does this PR solve?

nomasprime:
@janko Dispatch doesn't set the right compiler and therefore catch anything with `-I test` but it does with `-Itest`.

janko:
@nomasprime Ok, that makes sense. For others, the `rubyunit` compiler [has `ruby -Itest`](https://github.com/vim-ruby/vim-ruby/commit/5bf6bda1ef660bdf595625a01268ec532b2e77f6#diff-81bec943204a76cb74e7d0acd48e7db0) as one of the `makeprg`s, and dispatch.vim matches on those:

```
:echo dispatch#compiler_for_program('ruby -I test')
ruby

:echo dispatch#compiler_for_program('ruby -Itest')
rubyunit
```

PR #429, opened 2020-02-09
Fixes issues with running any kind of test with MacVim

willfish:
### What?

vim-test checks to see whether colours should be enabled in test output using the 'test#base#no_colors' function.

This function short-circuits when the current version of vim doesn't have the gui_running feature enabled (I imagine this is the majority of cases for users of vim (?)) 

```vim
function! test#base#no_colors() abort
  let strategy = get(g:, 'test#strategy', 'basic')
  return has('gui_running') && strategy ==# 'basic'
endfunction
```

I'm currently using [MacVim](https://github.com/macvim-dev/macvim) for a trial and am hitting the second check (i.e. `strategy ==# 'basic'`) since `has('gui_running')` is true.

However the strategy returned by the following call:
```vim
 let strategy = get(g:, 'test#strategy', 'basic')
```
is not a string and is in fact a dictionary object containing all of the strategies defined for the available run types (i.e. file, nearest and suite):
```vim
{'file':'iterm', 'nearest': 'iterm', 'suite': 'iterm'}
```

This PR fixes the failure arising from comparing a dictionary to a string:

image

### Why?

So that I can use vim-test with MacVim without it exploding :).

### A/C

### DIFF
--- a/autoload/test.vim
+++ b/autoload/test.vim
@@ -81,7 +81,7 @@ function! test#execute(runner, args, ...) abort
   call filter(args, '!empty(v:val)')
 
   let executable = test#base#executable(a:runner)
-  let args = test#base#build_args(a:runner, args)
+  let args = test#base#build_args(a:runner, args, strategy)
   let cmd = [executable] + args
   call filter(cmd, '!empty(v:val)')
 
--- a/autoload/test/base.vim
+++ b/autoload/test/base.vim
@@ -44,7 +44,7 @@ endfunction
 
 function! test#base#no_colors() abort
   let strategy = get(g:, 'test#strategy', 'basic')
-  return has('gui_running') && strategy ==# 'basic'
+  return has('gui_running') && strategy['file'] ==# 'basic'
 endfunction
 
 " Takes a position and a dictionary of patterns, and returns list of strings
--- a/autoload/test/base.vim
+++ b/autoload/test/base.vim
@@ -30,8 +30,8 @@ function! test#base#executable(runner) abort
   endif
 endfunction
 
-function! test#base#build_args(runner, args) abort
-  return test#{a:runner}#build_args(a:args)
+function! test#base#build_args(runner, args, strategy) abort
+  return test#{a:runner}#build_args(a:args, a:strategy)
 endfunction
 
 function! test#base#file_exists(file) abort
@@ -42,9 +42,8 @@ function! test#base#escape_regex(string) abort
   return escape(a:string, '?+*\^$.|{}[]()')
 endfunction
 
-function! test#base#no_colors() abort
-  let strategy = get(g:, 'test#strategy', 'basic')
-  return has('gui_running') && strategy['file'] ==# 'basic'
+function! test#base#no_colors(strategy) abort
+  return has('gui_running') && a:strategy ==# 'basic'
 endfunction
 
 " Takes a position and a dictionary of patterns, and returns list of strings
--- a/autoload/test/clojure/fireplacetest.vim
+++ b/autoload/test/clojure/fireplacetest.vim
@@ -22,7 +22,7 @@ function! test#clojure#fireplacetest#build_position(type, position) abort
   endif
 endfunction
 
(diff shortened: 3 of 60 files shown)


willfish:
I wanted to do a couple of things:

1. Understand what 'basic' should be compared against (i.e. for which run type - nearest, file or suite) - I'd like there to be a strategy available for the current run if at all possible.
2. Mock out the `has('gui_running')` functionality so that I can check the behaviour of this method in a base spec file to avoid this happening again in future.

codeinabox:
Does this behaviour only occur with MacVim?

willfish:
For me, yeah. I don't think neovim/vim return 1 for `has('gui_running')` in any circumstance and I'm not familiar with other graphical vim implementations

willfish:
I think I just answered question 1 above. I'm now passing the run strategy to the build_args for each runner which means the checks will work with MacVim now :).

willfish:
I appreciate that it ended up needing an interface change so seems bigger than it actually is.

codeinabox:
I've just downloaded MacVim to see if I can recreate the issue and when I run `echo get(g:, 'test#strategy', 'basic')` I just get a string not a dictionary. Are you assigning `test#strategy` anywhere?

willfish:
This is my configuration, yeah:

```vim
"""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""
" => vim-test
"""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""
if has('gui_macvim')
  let test#strategy = {
        \ 'nearest': 'iterm',
        \ 'file':    'iterm',
        \ 'suite':   'iterm',
        \}
else
  let test#strategy = {
        \ 'nearest': 'neovim',
        \ 'file':    'dispatch',
        \ 'suite':   'dispatch',
        \}
endif

let test#ruby#rspec#executable = "bundle exec rspec"

nnoremap <silent> <Leader>x :TestNearest<CR>
nnoremap <silent> <Leader>t :TestFile<CR>
nnoremap <silent> <Leader>r :TestSuite<CR>
nnoremap <silent> <Leader>e :TestLast<CR>
nnoremap <silent> <Leader>l :TestVisit<CR>
```

willfish:
I was following the documentation for using different strategies: https://github.com/janko/vim-test/commit/a8fc25c143d4b9417a744cdf3af81120d6390e22

codeinabox:
Thank you for your work on this and your patience. Seems reasonably sound though I wonder if there is a way to avoid passing the strategy to every runner's `build_args` - @janko do you know if this is possible?

willfish:
No worries codeinabox :). We could get more MacVim users using vim-test \0/

willfish:
Any thoughts on merging this @janko? Is really just a bug fix for a certain combination of config and MacVim.

Doesn't break/change the interface in a way that hasn't been updated/validated by running the tests.

janko:
This change will break any custom-made runners that people have created locally (because the `build_args` function is now given two arguments). I also don't like the fact that this introduces coupling between runners and strategies, because now runners would be able to generate different parameters based on the strategy. I know runners are not doing it at the moment, but this change would allow it, and that would introduce a slipper slope.

It would be better if we first determined whether colors should be generated, and then called `build_args` with that boolean value as the second argument (but we would also handle `build_args` accepting only one argument for backwards compatibility). That way the runner knows only what it needs to know – whether it can use colors.

willfish:
So AFAIK inside of the `test#base#no_colors` function we can do one of:

1. Pass in the run **type** and draw on a function that pulls out the strategy config for the specific run type (breaks custom runners)
2. Pass in the **strategy** (breaks custom runners)
3. Set/unset a variable that is in lexical scope of the entire run which specifies the type and/or strategy (doesn't break custom runners as no change in `test#<language>#<runner>#build_args` is required). I'm guessing this would be us using a global variable :scream: or equivalent?

How does the global variable approach sound @janko?

willfish:
This was superseded by https://github.com/janko/vim-test/pull/429

PR #447, opened 2020-03-06
Fix asyncrun_background global variable path

webdavis:
I apologize for introducing the breakage. Should I add a test for this? @codeinabox

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,27 +80,27 @@ Test.vim can run tests using different execution environments called
 let test#strategy = "dispatch"
 ```
 
-| Strategy                        | Identifier                       | Description                                                                                                |
-| :-----:                         | :-----:                          | :----------                                                                                                |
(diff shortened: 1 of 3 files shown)


webdavis:
- Hard-coded quickfix mode with `-mode=async`, otherwise `asyncrun_background` will break when `g:asyncrun_mode = 'term'`.
- Abstracted out the call to the `AsyncRunStop` to get ready for `asyncrun_background_term` support which should be finished soon.

webdavis:
Added an additional commit for `asyncrun_background_term` support.

webdavis:
Thanks for clarifying. Okay now it should be all good.

webdavis:
Changed `echom` to `echo`. When multiple tests are run in succession the global will be overwritten and the message buffer will not accurately reflect the order of tests. If the user wants to keep a message log then he/she should opt for `asyncrun_background_term` mode to catalog tests using tabs.

PR #449, opened 2020-03-12
Fix neovim strategy focusing terminal buffer

### LINKED ISSUE #448: Neovim terminal mode
Hi,

When I run my tests using the `neovim` strategy, it opens a new terminal but auto focus on the buffer. Then I have to press `<C-o>` to not close the buffer and go back to my split pane.

Is there a way to have it open but not switch my focus to the newly opened pane automatically?

chinleung:

This PR fixes #448.

The issue was `startinsert` which causes the insert mode in terminal and then you have to go back to your previous buffer manually.

To fix this, I've patched the strategy to have the same behaviour if when running a filtered test (`--filter`), otherwise, automatically go back to previous buffer.

The send the keys `G` to follow the scrolling of the terminal buffer and `<C-W><C-P>` to switch back to the last window.

The reason why I'm leaving the old behaviour for filtered test is because you expect a quick result when running a single test, so you don't want to manually change back to the terminal buffer. When running a suite or file, the tests could take longer, hence going back to your previous buffer to continue your work.

### DIFF
--- a/autoload/test/strategy.vim
+++ b/autoload/test/strategy.vim
@@ -76,7 +76,13 @@ function! test#strategy#neovim(cmd) abort
   execute term_position . ' new'
   call termopen(a:cmd)
   au BufDelete <buffer> wincmd p " switch back to last window on delete
-  execute feedkeys("G\<C-W>\<C-P>") " switch back to last window
+
+  " Start insert mode only when running a specific test
+  if stridx(a:cmd, '--filter') != -1
+      startinsert
+  else
+    execute feedkeys("G\<C-W>\<C-P>")
+  endif
 endfunction
 
 function! test#strategy#vimterminal(cmd) abort
--- a/autoload/test/strategy.vim
+++ b/autoload/test/strategy.vim
@@ -75,8 +75,8 @@ function! test#strategy#neovim(cmd) abort
   let term_position = get(g:, 'test#neovim#term_position', 'botright')
   execute term_position . ' new'
   call termopen(a:cmd)
-  au BufDelete <buffer> wincmd p " switch back to last window
-  startinsert
+  au BufDelete <buffer> wincmd p " switch back to last window on delete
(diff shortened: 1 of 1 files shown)


autoload/test/strategy.vim
```diff
@@ -75,8 +75,14 @@ function! test#strategy#neovim(cmd) abort
   let term_position = get(g:, 'test#neovim#term_position', 'botright')
   execute term_position . ' new'
   call termopen(a:cmd)
-  au BufDelete <buffer> wincmd p " switch back to last window
-  startinsert
+  au BufDelete <buffer> wincmd p " switch back to last window on delete
+
+  " Start insert mode only when running a specific test
+  if stridx(a:cmd, '--filter') != -1
```
janko: This logic is runner-specific, so it doesn't belong in a strategy. I understand the need for it, but we have to find another way.

chinleung: Oh yeah that's true. Forgot about the other runners. Is there a way to differentiate the command that was ran? Like detect if it's `:TestNearest`?

janko: Typically you'd set different CLI options depending on whether it was `:TestNearest`, `:TestFile` or `:TestSuite`, but in this case we want different _strategy_ options, and we don't yet have a concept of "strategy options" that you could pass via the command.

At the moment strategies don't have the knowledge whether `:TestNearest` was used, they receive only the test command. I'm not sure whether giving strategies this knowledge is the correct thing to do, but maybe it would make sense.

An alternative that's currently possible is to use different strategies for different granularity. That would mean we'd need to have two different strategies, and then the user can configure which one they want for which granularity.

chinleung: Okay guess we can leave it like the way it used to be until there's a concept of strategy options. Not sure if I'm the only one who needs the no auto focus update anyways, so I'll just use these changes locally.

PR #451, opened 2020-03-14
Mocha flexible file pattern

codeinabox:
I am working on a project where Mocha unit tests aren't in the test directory which is suggested but not required. This updates the file pattern to handle this and also generates a glob pattern for the TestSuite


codeinabox:
@rumblesan do you have any feedback on this? I remember you helped out improve the Mocha runner

PR #452, opened 2020-03-19
Fix coloring flag not working on GUI with granular strategies

janko:
The `test#base#no_colors()` function was looking at the global strategy variable, which can be a dictionary when per-granularity strategies was specified. This function doesn't know the granularity, since the `build_args()` runner function is called from `test#execute()`, which doesn't have granularity knowledge (since that function can be called directly via runner commands such as `:RSpec`).

Instead of requiring runners to call into our function for determining whether coloring can be used, we change the code to determine this before calling `build_args()`, and then just pass the boolean value into the function. At that point the strategy has already been resolved, so we can just use it right there and then.

I didn't find a way in VimScript to check how many arguments a function supports, so we had to resort to try/catch for backwards compatibility. We could have updated all runners to accept the second argument, but that wouldn't solve any custom-made runners that people might have created locally.

Closes #429

### DIFF
--- a/autoload/test/base.vim
+++ b/autoload/test/base.vim
@@ -30,8 +30,14 @@ function! test#base#executable(runner) abort
   endif
 endfunction
 
-function! test#base#build_args(runner, args) abort
-  return test#{a:runner}#build_args(a:args)
+function! test#base#build_args(runner, args, strategy) abort
+  let no_color = has('gui_running') && strategy ==# 'basic'
+
+  try
+    return test#{a:runner}#build_args(a:args, !no_color)
+  catch /^Vim\%((\a\+)\)\=:E118/ " too many arguments
+    return test#{a:runner}#build_args(a:args)
+  endtry
 endfunction
 
 function! test#base#file_exists(file) abort
@@ -40,11 +46,6 @@ endfunction
 
 function! test#base#escape_regex(string) abort
   return escape(a:string, '?+*\^$.|{}[]()')
-endfunction
-
-function! test#base#no_colors() abort
-  let strategy = get(g:, 'test#strategy', 'basic')
-  return has('gui_running') && strategy ==# 'basic'
 endfunction
 
 " Takes a position and a dictionary of patterns, and returns list of strings
--- a/autoload/test/base.vim
+++ b/autoload/test/base.vim
@@ -30,7 +30,7 @@ function! test#base#executable(runner) abort
   endif
 endfunction
 
-function! test#base#build_args(runner, args, strategy) abort
+function! test#base#build_args(runner, args, strategy)
   let no_color = has('gui_running') && a:strategy ==# 'basic'
 
   try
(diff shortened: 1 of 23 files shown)


janko:
@willfish This is what I had in mind ☝️ (though the `catch` blog doesn't work on CI at the moment)

autoload/test/base.vim
```diff
@@ -30,8 +30,14 @@ function! test#base#executable(runner) abort
   endif
 endfunction
 
-function! test#base#build_args(runner, args) abort
-  return test#{a:runner}#build_args(a:args)
+function! test#base#build_args(runner, args, strategy) abort
```
willfish:
```suggestion
function! test#base#build_args(runner, args, strategy)
```

Not super familiar with vim script, but do we need to kill the abort if we mean to use exceptions to control the branches of the function?

willfish:
This approach seems more robust to me, yeah 🎉

autoload/test/base.vim
```diff
@@ -30,8 +30,14 @@ function! test#base#executable(runner) abort
...
+  let no_color = has('gui_running') && strategy ==# 'basic'
```
willfish:
```suggestion
  let no_color = has('gui_running') && a:strategy ==# 'basic'
```

willfish:
With those changes it's working for me using MacVim and Dispatch and the multiple strategy config.

```vim
"""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""
" => vim-test
"""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""

let test#strategy = {
      \ 'nearest': 'dispatch',
      \ 'file':    'dispatch',
      \ 'suite':   'dispatch_background',
      \}

let test#ruby#rspec#executable = 'bundle exec rspec'

nnoremap <silent> <Leader>x :TestNearest<CR>
nnoremap <silent> <Leader>t :TestFile<CR>
nnoremap <silent> <Leader>r :TestSuite<CR>
nnoremap <silent> <Leader>e :TestLast<CR>
nnoremap <silent> <Leader>l :TestVisit<CR>
```

I tried nearest, file and suite and all work

image

willfish:
@janko Do you mind reviewing this again?

janko:
@willfish I appreciate the code review 🙏

I realized that I don't have time to maintain this project anymore, so I've moved the repo to a new `vim-test` organization, and have added you as a member (you should have push access).

You're free to get this pull request merged and do any other changes you want in the future 😉

willfish:
@janko The specs were failing on older versions of vim. I've upgraded and cached the one we build in travis. It doesn't add loads of time to the build and makes these specs pass. Let me know if you think this is okay and I'll merge this change :)

janko:
Looks great to me! 👍 And I really appreciate you picking this up ❤️

willfish:
No worries :).

PR #458, opened 2020-03-28
Tweak tmuxify strategy a bit

phongnh:
`tmuxify#pane_send_raw` send keys as single to pane, in some program
 like `less` it can understand these keys. However, tmux in copy mode
 does not understand it.

Please help me to review this PR. Thanks.


phongnh:
Thanks @codeinabox. Tried to test a bit, it works okay. You can check here https://asciinema.org/a/bc0m5n35ofJXqOFsYY5MOhmWz.

PR #461, opened 2020-04-09
fix cargotests run from /tests

meain:
Added support when there is no `mod` defined or if the mod name is `test` instead of `tests`

---

### DIFF
--- a/autoload/test/rust/cargotest.vim
+++ b/autoload/test/rust/cargotest.vim
@@ -5,7 +5,7 @@ endif
 if !exists('g:test#rust#cargotest#test_patterns')
   let g:test#rust#cargotest#test_patterns = {
         \ 'test': ['\v(#\[test\])'],
-        \ 'namespace': []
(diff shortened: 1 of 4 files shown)


meain:
Looks like we will have to check if it is `test` or `tests` based on the mod.

autoload/test/rust/cargotest.vim
```diff
@@ -61,28 +61,42 @@ function! s:nearest_test(position) abort
...
+    return ''
+  else
+    let l:modules = l:modules[1:]
+    if len(l:modules) > 0
+      return join(l:modules, '::') . '::'
+    else
+      return ''
+    endif
+  endif
+  " return join(l:modules, '::') . '::'
```
codeinabox, line 101: Please remove if no longer needed

meain: Woops, left it there accidentally. Fixed

PR #471, opened 2020-05-22
Use paratest as the phpunit executable if present

codeinabox:
This resolves #424 with a simpler solution of just using paratest as the phpunit runner executable if present


partounian:
LGTM 🚀

vivoconunxino:
when executing `TestNearest` now throws `Option --filter is not implemented for non functional mode` because of paratest.

codeinabox:
Thank you for spotting this. I'll see what what is possible but another option that might be simpler is adding `--functional` for TestNearest with paratest

PR #475, opened 2020-05-23
Update test#javascript#has_package to use json_decode if present

### LINKED ISSUE #474: Enhance JS package.json runner detection to use json_decode
The JS test runner detection which currently greps the package.json file could be improved by using the new `json_decode` feature in Vim. This is a reminder to myself to do this.

codeinabox:
Resolves #474

### DIFF
--- a/autoload/test/javascript.vim
+++ b/autoload/test/javascript.vim
@@ -4,7 +4,14 @@ let test#javascript#patterns = {
 \}
 
 function! test#javascript#has_package(package) abort
-  for line in readfile('package.json')
+  let l:packages = readfile('package.json')
+
+  if exists('*json_decode')
+	let l:dict = json_decode(packages)
+	return has_key(get(dict, 'dependencies', {}), a:package) || has_key(get(dict, 'devDependencies', {}),  a:package)
+  endif
+
+  for line in packages
     if line =~ '"'.a:package.'"'
       return 1
     endif
--- a/autoload/test/javascript.vim
+++ b/autoload/test/javascript.vim
@@ -7,7 +7,7 @@ function! test#javascript#has_package(package) abort
   let l:packages = readfile('package.json')
 
   if exists('*json_decode')
-	let l:dict = json_decode(packages)
+	let l:dict = json_decode(join(packages, ''))
 	return has_key(get(dict, 'dependencies', {}), a:package) || has_key(get(dict, 'devDependencies', {}),  a:package)
   endif
 


codeinabox:
@willfish @janko this works for me when testing manually but fails with vim-flavor and I am scratching my head as to why 🤔

willfish:
@codeinabox Which version of vim are you running?

willfish:
Using vvm locally I'm able to recreate the issue.

It seems that the latest version of vim from vim-org has broken the support for readfile outputs in the jsondecode function (which used to support lists).

### Before documentation

```help
json_decode({expr})					*json_decode()*
		Convert {expr} from JSON object.  Accepts |readfile()|-style
		list as the input, as well as regular string.  May output any
		Vim value. In the following cases it will output
		|msgpack-special-dict|:
		1. Dictionary contains duplicate key.
		2. Dictionary contains empty key.
		3. String contains NUL byte.  Two special dictionaries: for
		   dictionary and for string will be emitted in case string
		   with NUL byte was a dictionary key.

		Note: function treats its input as UTF-8 always.  The JSON
		standard allows only a few encodings, of which UTF-8 is
		recommended and the only one required to be supported.
		Non-UTF-8 characters are an error.
```


### After documentation

```
json_decode({string})					*json_decode()*
		This parses a JSON formatted string and returns the equivalent
		in Vim values.  See |json_encode()| for the relation between
		JSON and Vim values.
		The decoding is permissive:
		- A trailing comma in an array and object is ignored, e.g.
		  "[1, 2, ]" is the same as "[1, 2]".
		- Integer keys are accepted in objects, e.g. {1:2} is the
		  same as {"1":2}.
		- More floating point numbers are recognized, e.g. "1." for
		  "1.0", or "001.2" for "1.2". Special floating point values
		  "Infinity", "-Infinity" and "NaN" (capitalization ignored)
		  are accepted.
		- Leading zeroes in integer numbers are ignored, e.g. "012"
		  for "12" or "-012" for "-12".
		- Capitalization is ignored in literal names null, true or
		  false, e.g. "NULL" for "null", "True" for "true".
		- Control characters U+0000 through U+001F which are not
		  escaped in strings are accepted, e.g. "	" (tab
		  character in string) for "\t".
		- An empty JSON expression or made of only spaces is accepted
		  and results in v:none.
		- Backslash in an invalid 2-character sequence escape is
		  ignored, e.g. "\a" is decoded as "a".
		- A correct surrogate pair in JSON strings should normally be
		  a 12 character sequence such as "\uD834\uDD1E", but
		  json_decode() silently accepts truncated surrogate pairs
		  such as "\uD834" or "\uD834\u"
								*E938*
		A duplicate key in an object, valid in rfc7159, is not
		accepted by json_decode() as the result must be a valid Vim
		type, e.g. this fails: {"a":"b", "a":"c"}

		Can also be used as a |method|: >
			ReadObject()->json_decode()
```

I fixed this locally by using `join(packages, '')` and passing the result of that to the `jsondecode`.

This works with both the new and the older versions of jsondecode since vim 7.4

autoload/test/javascript.vim
```diff
@@ -4,7 +4,14 @@ let test#javascript#patterns = {
 \}
 
 function! test#javascript#has_package(package) abort
-  for line in readfile('package.json')
+  let l:packages = readfile('package.json')
+
+  if exists('*json_decode')
+	let l:dict = json_decode(packages)
```
willfish, line 10:
```suggestion
        let l:joined_packages = join(packages, '')
	let l:dict = json_decode(joined_packages)
```

willfish, line 10:
```suggestion
	let l:dict = json_decode(join(packages, ''))
```

codeinabox:
Thank you for the help @willfish. I had tested against Neovim which doesn't have this issue.

PR #480, opened 2020-05-30
Bugfix for ParaTest nearest

codeinabox:
This fixes the following issue:

> when executing `TestNearest` now throws `Option --filter is not implemented for non functional mode` because of paratest.
 
 _Originally posted by @vivoconunxino in https://github.com/vim-test/vim-test/pull/471#issuecomment-636066013_


codeinabox:
@vivoconunxino could you please test this branch and let me know how you get on, it will run paratest in functional mode when there is a filter.

vivoconunxino:
@codeinabox tested and working :-)

Issue #272, opened 2018-01-31
Package.json detection fails if the package.json is not in the root


entropitor:
E.g. I'm working in a mono-repo with multiple package.json's in my repository, vim-test cannot find the other ones. `vim-lsp` for examples can cope with this using some helper functions

`        \ 'root_uri':{server_info->lsp#utils#path_to_uri(lsp#utils#find_nearest_parent_file_directory(lsp#utils#get_buffer_path(), '.flowconfig'))},
`

It would be nice if we could also configure this for vim-test

codeinabox:
I guess it could be possible to have an option to manually override the path it checks for package.json but in this case are you staying there are multiple ones, and I am guessing each of them might have a different test runner?

I need to check with @janko-m but I believe the scope of this plugin is just for it running in the root directory of a single project

janko-m:
I'm not sure I understand yet what exactly vim-test needs to do here, but if any feature request that would otherwise require overriding the `:Test*` commands I would consider to be in the scope of vim-test.

If this use case is too specific, it might also be worthwhile to add some Vim auto-commands to allow people to change some things dynamically.

entropitor:
@codeinabox The way `vim-lsp` does things is by determining the command to run based on which buffer you have open. So potentially that could mean multiple test runners, however it's always determined which one it should be (by just going up until you find a package.json)

codeinabox:
Yes but there will be package.json in people's vendor directories

entropitor:
@codeinabox But that is not in the path to the file, is it? It's normally in a `node_modules` directory next to the app, I cannot imagine any situation in which a vendor's package.json would be found anywhere on the path from the file to the root directory

codeinabox:
Now I am working on a project with mono repo I can see the problem but I don't see this a problem specifically with vim-test but any plugin that assumes the current working directory is the project directory. Keen to find a solution but ideally one that works across multiple plugins and not just vim-test.

entropitor:
@codeinabox Given this project already uses projectionist, maybe it should use the result of the function call `call projectionist#path()` as the project dir for a file?

codeinabox:
Interesting, so projectionist let's you define multiple projects and `call projectionist#path()` will give you the project root path for that file?

entropitor:
Yes, exactly!

codeinabox:
> @codeinabox Given this project already uses projectionist, maybe it should use the result of the function call `call projectionist#path()` as the project dir for a file?

[...]

PR #492, opened 2020-07-15
Make a note about working with a monorepo

codeinabox:
Given all the recent issues when working with a monorepo and #272 I feel it's best to add this limitation to the documentation and explain how users can mitigate this.


codeinabox:
I've since noticed that the documentation already makes mention of this so have just added the word monorepo

PR #502, opened 2020-07-31
Only set --functional for nearest if the executable after overrides is paratest

### LINKED ISSUE #501: phpunit - when paratest is installed, all other executables break when using TestNearest
#480 introduced code so that when `paratest` is detected (by checking for the presence of `./vendor/bin/paratest`) it prepends `--functional ` to the test name, regardless of whether `paratest` is the executable.

https://github.com/vim-test/vim-test/blob/master/autoload/test/php/phpunit.vim#L28-L30

This causes other executables to choke with an `unrecognized option --functional`

In my case, `paratest` is installed to make CI builds faster because it runs multiple tests in parallel. However for local development it can make debugging harder because it seems to swallow `stdout` so things like `dd()` can't be used. I would prefer to use standard `phpunit` (or the improved output of `php artisan test` as mentioned in #445), especially for `TestNearest`, but currently this code is making that not possible.

I'm happy to submit a PR for this, but I'm not familiar enough with the architecture of this (or vimscript) to know the best way to prevent the bug that #480 solved, while also allowing different executables to be used when `paratest` is present.

codeinabox:
Resolves #501

### DIFF
--- a/autoload/test/php/phpunit.vim
+++ b/autoload/test/php/phpunit.vim
@@ -25,7 +25,7 @@ function! test#php#phpunit#build_position(type, position) abort
   if a:type ==# 'nearest'
     let name = s:nearest_test(a:position)
     if !empty(name) | let name = '--filter '.shellescape('::'.name, 1) | endif
-    if !empty(name) && filereadable('./vendor/bin/paratest')
+    if !empty(name) && test#php#phpunit#executable() =~ 'paratest'
       let name = '--functional '. name
     endif
     return [name, a:position['file']]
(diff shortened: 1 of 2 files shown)


jessarcher:
This works perfectly for me. Thank you!

In a project with paratest installed, I can now override the executable to be `php artisan test` or `phpunit` and it no longer adds `--functional`, so the test runner doesn't fail.

I also tested that it continues to uses paratest by default and continues to use `--functional` for `TestNearest` (so no regressions). I also tried explicitly setting the executable to `paratest` and `./vendor/bin/paratest` and the `--functional` arg was always added appropriately.

PR #508, opened 2020-08-14
Fix TAP nearest to support multiline match

codeinabox:


codeinabox:
@janko need a little bit of help on this. I tried modifying the nearest pattern for TAP to look for new lines using `\_` like below but still had issues

```viml
  let patterns = {
        \ 'test': ['\v^\_s*%(%(%(tap|t)\.)?test)\_s*[( ]\_s*%("|''|`)(.*)%("|''|`)'],
        \ 'namespace': [],
        \}
```

codeinabox:
@janko I know understand why this isn't working, it's because `test#base#nearest_test_in_line` goes through line by line which means that multi-line matching for test nearest isn't currently possible. Any suggestions on how to proceed with this?

janko:
Not sure, line-by-line matching is needed to detect nesting. I don’t know how to handle multiline matching as well 🤔

codeinabox:
What if test nearest match started at a single line and then kept expanding in range in both directions till a match was found? For example I run it at line 12 and it searches that line, then tries lines 11 to 13, then 10 to 14 and so forth.

janko:
I don’t know exactly how you would calculate indentation with regexes. Currently the code updates some state after each line. And it’s only going up from the current line IIRC.

PR #513, opened 2020-08-26
Add notes on PHPUnit alternate runners

codeinabox:
This adds some notes on PHPUnit alternate runners, something I meant to do after addressing #501


codeinabox:
@jessarcher and @rodrigore could you please sanity check this documentation makes sense

jessarcher:
Looks good to me! Only tiny comment is that the ruby section is using code blocks instead of inline code, so I'm not sure if this should match the same style?

PR #534, opened 2020-12-12
Update Rails detection to look for application.rb

BlakeWilliams:
Currently the Rails detection logic looks for the presence of `rails` in
the `Gemfile.lock` file. This works in most cases, but is incorrectly
detected when Rails is a dependency of a gem.

This updates the Rails detection code to look for
`config/application.rb` which should be present in every Rails
application, but not in projects where Rails is only a dependency.


codeinabox (approved):
👍  Thank you for this

PR #537, opened 2021-01-02
Fix cargo test invocation for main.rs

alexgenco:
Before, if you ran tests in `main.rs`, it would invoke `cargo test main::` and not find any tests, but I don't think typically you will actually have a `main` module in `main.rs`.

- [x] Update the README accordingly (n/a)
- [x] Update the Vim documentation in `doc/test.txt` (n/a)


codeinabox (approved):
Thank you for this!

PR #543, opened 2021-01-20
Make Python3 the default Python executable

fdelacruz:


codeinabox:
Thank you for raising this PR. I do intend to improve the logic for Python runners so it detects the version the correct executable so please do let me know if you have any suggestions.

codeinabox:
A heads-up that I've added logic #750 to detect the Python executable

PR #544, opened 2021-01-20
Update PyUnitTest to use Python3

fdelacruz:


fdelacruz:
These tests are failing because the "Actual value" has python. Even if, for whatever reason, this PR is not accepted. Can that be changed?
Screen Shot 2021-01-20 at 3 04 22 PM

codeinabox:
Thank you for this PR, could you include the changes from #543 and that will make the tests in this PR pass

fdelacruz:
These tests fail because they seem to be fixated to use python not python3. I don't see where is that setup, its not in the .travis.yml

fdelacruz:
These tests fail because they seem to be set to use python not python3. I don't see where is that configured, since it's not in the .travis.yml

Issue #505, opened 2020-08-07
Jest tests failing


withrocks:
Jest tests are failing in 007972e2 with this error:

```
Error detected while processing function test#run[26]..test#execute[18]..test#base#build_args:
line    4:
E118: Too many arguments for function: test#javascript#jest#build_args
E15: Invalid expression: test#{a:runner}#build_args(a:args, !no_color)
Error detected while processing function test#run:
line   26:
E171: Missing :endif
```

Running bisect shows that this was introduced in 19b84e68.

I think I might be able to fix it myself, but want to start by submitting an issue in any case.

withrocks:
Ok, I think I might have an inkling why this happens. The commit that introduced this added this catch to ensure that what I'm experiencing doesn't happen: https://github.com/vim-test/vim-test/blob/19b84e68/autoload/test/base.vim#L38

I guess that this catch doesn't work correctly in NeoVim (which I'm using). I'll look into what needs to happen in that case.

withrocks:
@janko commented in the commit on this behaviour:

```
I didn't find a way in VimScript to check how many arguments a
function supports, so we had to resort to try/catch for backwards
compatibility. We could have updated all runners to accept the second
argument, but that wouldn't solve any custom-made runners that people
might have created locally.
```

I wonder if it would be a good idea to have all runners included in the package support the second argument and also introduce a new function, e.g. called `interface_version` to make dealing with breaking changes easier. If a runner script has that function, it should be called and used to decide which function to call. If it doesn't have, interface_version `1` could be assumed. This could be used for any other function that requires breaking changes too and I believe it might be cleaner (but more code is involved).

The code might look something like this (notice that I'm a vimscript noob):

```
" In base.vim:

" Returns the interface version of the test runner script. If not provided,
" version 1 is assumed. This is so breaking changes can be introduced.
function! test#base#runner_interface_version(runner)
  if exists('g:test#'.a:runner.'interface_version')
    return test#{a:runner}#interface_version()
  else
    return 1
  endif
endfunction

function! test#base#build_args(runner, args, strategy)
  let no_color = has('gui_running') && a:strategy ==# 'basic'
  let interface_version = test#base#runner_interface_version(a:runner)

  if interface_version == 1
    return test#{a:runner}#build_args(a:args)
  else
    return test#{a:runner}#build_args(a:args, !no_color)
  endif
endfunction
```

[...]

PR #545, opened 2021-01-25
Avoid uncaught exception in test#base#build_args()

dmitrivereshchagin:
Test runners that provide build_args() taking one argument cannot be used in Vim before 8.0.1423 (and Neovim, #505).

This change adds workaround for Vim (and Neovim) versions that fail to caught an exception thrown from return statement (see vim/vim#2483).


davycardinaal:
I ran into the same issue, thanks for this fix! Please merge :-)

jordyvanvorselen:
Thanks! Was running into this as well.

codeinabox (approved):

PR #546, opened 2021-01-27
Minitest: don't pick up RSpec files by mistake

fallwith:
I'm using (and loving, thank you!) vim-test in an RSpec based Ruby project. Everything was working great until I started trying to test a `Test` class. Because my class name is `Test`, the corresponding RSpec file name is `test_spec.rb`. When I attempt to run the RSpec tests inside this file, vim-test attempts to run Minitest instead of RSpec. This is understandable, as usually Minitest files have `test` in the name whereas RSpec files have `spec` in the name, but in this case my valid RSpec file has both `test` and `spec`.

This PR updates the Minitest file pattern regex to ignore a file starting with 'test' if it also ends with 'spec.rb'. I've added a new test fixture and unit test and the existing unit tests still pass.

- [x] (N/A) Update the Vim documentation in `doc/test.txt`

### DIFF
--- a/README.md
+++ b/README.md
@@ -370,7 +370,7 @@ certain testing framework. You can override that pattern by overriding the
 `file_pattern` variable:
 
 ```vim
-let test#ruby#minitest#file_pattern = '_spec\.rb' " the default is '(((^|/)test_.+)|_test)\.rb'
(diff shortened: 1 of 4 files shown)


codeinabox (approved):
Thank you for this and your kind praise 🙏

PR #547, opened 2021-02-15
Add support for Laravel Sail

michaeldyrynda:

When using [Laravel Sail](https://laravel.com/docs/sail), PHPUnit tests need to be passed through to Sail within Docker, in order to run correctly, as PHP is likely not installed on the host and/or access to the Docker guest may not be possible.

This just checks if `./vendor/bin/sail` exists and uses it as the runner.


codeinabox (approved):
This is great, thank you for the addition 🥳

halivert:
¿How does this works with vim-dispatch? 
If sail needs to be in a TTY

PR #552, opened 2021-03-23
Run Dart tests with `pub run test`

brianegan:
Thanks for the cool plugin :) I wanted to update it a bit to work for regular old dart packages, not just those fancy Flutter apps.

  * Plain Old Dart Packages run with the default `pub run test`. 
  * Flutter apps that make use of the `flutter_test` package run with `flutter test`.

- [X] Update the README accordingly
- [X] Update the Vim documentation in `doc/test.txt`


codeinabox (approved):
A great addition, thank you for this

akinsho:
@brianegan the standalone `pub` command is deprecated see [here](https://dart.dev/tools/pub/cmd), because of that this change breaks for anyone not using the global pub command which will be the case for anyone who has dart installed as part of installing flutter. Note that `dart test` works and I think is now the recommended way of running dart tests not just for flutter (also discussed [here](https://github.com/dart-lang/test/issues/1298))

codeinabox:
Shall we revert this change or would someone like to some logic to check for the presence of `pub` before invoking it?

akinsho:
Would be good to get @brianegan's input. Since I'm not really familiar with what changes will be needed and if just dropping `dart test` in everywhere that `pub run test` is used will be enough. Seems from the docs like it would be but I use dart entirely in the context of flutter so I could easily be missing something.

Maybe we can leave this for a day or 2 to see if Brian or I get to trying to raise a PR to use `dart test` instead and if neither of us do then revert this. I think the number of users who are probably affected by this change might be greater than the people who were aware of this addition *and* are using the global `pub run test`

brianegan:
Hey, sorry about that -- I almost always use the standalone version of Dart for normal Dart projects, and the Flutter version of Dart only for Flutter projects, so I hadn't run into this issue.

I've opened up the following PR as a fix: https://github.com/vim-test/vim-test/pull/552

brianegan:
Hm, Can't edit my comment. Correct new PR is: https://github.com/vim-test/vim-test/pull/556

PR #553, opened 2021-03-24
Update floaterm strategy

voldikss:

----

Right now, floating terminal window will be closed as soon as tests finished if the floaterm's global option `g:floaterm_autoclose` is `1` or `2`. This is not good design since user should be able to see the testing result.

Hence, the PR aims to keep the window opened until user move the cursor.

cc @codeinabox


codeinabox (approved):
Thank you

mhanberg:
Is there a user configureable way to undo this change? I specifically set autoclose to 1 so that passing tests will close the window immediately but failing tests stay open.

PR #562, opened 2021-04-27
WIP: add testthat runner

gegnew:
Heyo! I've starting trying to implement a test runner for `R` using `testthat`. It's proving to be a bit nontrivial, because there is no `testthat` executable that accepts flag arguments (mostly). To run `testthat` from the command line, you have to use `Rscript`, and pass valid R code as a string to the "-e" flag.

i.e.:
```bash
Rscript -e "library(testthat); test_local()"
```

- I've managed to implement TestFile. 
- TestSuite alone is no problem, but I haven't managed to implement it alongside TestFile. The test executable would ideally be `Rscript -e "library(testthat); test_local()"`, but `test#r#testthat#build_position` must handle the `test_local` part of the script like: `"test_local(filter='something')"`, since the filter is not a flag.
- I have no idea how to implement TestNearest; it may not be possible.

I'd love some help, if anyone has time!


codeinabox:
Thank you for this and your patience. Would love to help out with this, do you have an example of the scenario you are trying to address?

gegnew:
I'm trying to address the situation where I'm editing R in vim (instead of Rstudio), and would like to run `testthat` files from vim. Something as simple as this test:

```r
# my-r-file.R
test_that("is equal", {
    expect_equal(1, 1)
})
```

autoload/test/r/testthat.vim
```diff
@@ -0,0 +1,49 @@
+if !exists('g:test#r#testthat#file_pattern')
+  let g:test#r#testthat#file_pattern = '\v(test-[^/]+|[^/]+)\.R$'
```
pingortle, line 2: Is this regex case sensitive? If so, I'd like to support lowercase `.r` files as well because case insensitive filesystem users often play fast and loose with casing in filenames. (I myself am guilty of this ✋)

```suggestion
  let g:test#r#testthat#file_pattern = '\v(test-[^/]+|[^/]+)\.[Rr]$'
```

PR #565, opened 2021-05-03
Made maven java dir optional

elevin72:
Very minor (and probably niche) change. Made the maven directory src/main/java optional such that src/main will also match.
Why would I want this? I hate deep directory structures with no meaning, and since my project is excursively java it was redundant. This tiny regex change shouldn't effect anyone with a normal maven structure, and will make my weird structure compatible with the plugin.

### DIFF
--- a/autoload/test/java/maventest.vim
+++ b/autoload/test/java/maventest.vim
@@ -50,7 +50,7 @@ function! s:get_java_package(filepath)
   let abspath = substitute(a:filepath, '\\', '/', 'g')
 
   " strip path-to-project and maven-boilerplate dir-structure
-  let relpath = substitute(abspath, '^.*src/\(main\|test\)/\(java/\)?', "", "g")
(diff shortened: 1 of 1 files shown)


codeinabox (commented):
Thank you for this. Could you possibly add a spec for this scenario? Happy to lend a hand with this

elevin72:
> Thank you for this. Could you possibly add a spec for this scenario? Happy to lend a hand with this

I'm not sure what you mean by this. Could you explain what you're looking for?

codeinabox:
> > Thank you for this. Could you possibly add a spec for this scenario? Happy to lend a hand with this
> 
> I'm not sure what you mean by this. Could you explain what you're looking for?

Adding to `spec/maventest_spec.vim` and the fixtures in `spec/fixtures/maven/` with an example of this scenario

codeinabox:
Realised the existing tests are already sufficient so happy to merge this in

PR #568, opened 2021-05-18
Replace rails check with railties check

BlakeWilliams:
Some applications don't use the `rails` gem explicitly and instead
depend on its components separately, such as `activerecord`,
`activesupport`, etc. When this is how an application requires rails, it
causes the test command to use `bin/rake` instead of `bin/rails` causing
tests not to run.

This resolves the issue by checking the `railties` version which should
always be present, even when `rails` isn't.

### DIFF
--- a/autoload/test/ruby/rails.vim
+++ b/autoload/test/ruby/rails.vim
@@ -45,7 +45,7 @@ endfunction
 function! s:rails_version() abort
   if filereadable('Gemfile.lock')
     for line in readfile('Gemfile.lock')
-      let version_string = matchstr(line, '\v^ *rails \(\zs\d+\.\d+\..+\ze\)')
+      let version_string = matchstr(line, '\v^ *railties \(\zs\d+\.\d+\..+\ze\)')
       if version_string
         break
       endif
--- a/spec/fixtures/rails-engine/Gemfile.lock
+++ b/spec/fixtures/rails-engine/Gemfile.lock
@@ -1,1 +1,2 @@
-rails (5.2.3)
+railties (>= 4.2.0)
+railties (5.2.3)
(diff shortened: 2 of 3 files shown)


codeinabox (commented):
Will this break things for legacy projects that have a Gemfile.lock containing `rails (5.2.3)` instead of `railties`?

BlakeWilliams:
> Will this break things for legacy projects that have a Gemfile.lock containing `rails (5.2.3)` instead of `railties`?

Great question, I'm not sure but I think it should continue working. I'll see if I can track down a project with that version and test it out (I'm running this patch locally so I can run tests).

codeinabox:
> > Will this break things for legacy projects that have a Gemfile.lock containing `rails (5.2.3)` instead of `railties`?
> 
> Great question, I'm not sure but I think it should continue working. I'll see if I can track down a project with that version and test it out (I'm running this patch locally so I can run tests).

Thank you. The other option is to add to the spec and fixtures with that scenario and make sure the behaviour still works.

BlakeWilliams:
> Thank you. The other option is to add to the spec and fixtures with that scenario and make sure the behaviour still works.

Totally, I was curious what a `Gemfile.lock` would look like generated from an older version of Rails. I assumed they would include `railties` and it looks like that's the case.

Here's the generated `Gemfile.lock` for `5.2.6` which includes railties: https://gist.github.com/BlakeWilliams/863ef0e96e588fd1796f9016d09668d0

Happy to add back the `rails (5.2.3)` line to the fixture if preferred though!

codeinabox:
> > Thank you. The other option is to add to the spec and fixtures with that scenario and make sure the behaviour still works.
> 
> Totally, I was curious what a `Gemfile.lock` would look like generated from an older version of Rails. I assumed they would include `railties` and it looks like that's the case.
> 
> Here's the generated `Gemfile.lock` for `5.2.6` which includes railties: https://gist.github.com/BlakeWilliams/863ef0e96e588fd1796f9016d09668d0
> 
> Happy to add back the `rails (5.2.3)` line to the fixture if preferred though!

Thank you for the confirmation, happy with this change as is

codeinabox (approved):

PR #578, opened 2021-08-03
Allow starting neovim term in normal mode

### LINKED ISSUE #577: Allow opening neovim :terminal in normal mode
The current behaviour is running tests with the `neovim` strategy opens a terminal in insert mode. This makes it easy to close the terminal by pressing a key, but it means you need to switch to normal mode `<C-\><C-n>` to navigate.

I would prefer to be able to have the terminal open in normal mode, so that I can navigate the test's output, and use my other defined normal mode commands (such as split navigation) more easily. Is this something that is currently possible, or could it be introduced as an option?

DanielPower:
Resolves #577 

- [x] ~Add fixtures and spec when implementing or updating a test runner~

Added new options `test#neovim#start_normal` and `test#basic#start_normal` which allow not switching to insert mode when opening a terminal in Neovim.

### DIFF
--- a/README.md
+++ b/README.md
@@ -145,6 +145,14 @@ if has('nvim')
 endif
 ```
 
+If you prefer, you can instead have the terminal open in normal mode, so it does
+not close on key press.
+
+```vim
+let g:test#neovim#start_normal = 1 # If using neovim strategy
+let g:test#basic#start_normal = 1 # If using basic strategy
+```
+
 By default vim-test will echo the test command before running it. You can
 disable this behavior with:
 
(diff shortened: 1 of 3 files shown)


DanielPower:
Two questions I would ask of reviewers are:
1. What do you think of `start_normal` as the name of the variable for this? Is there a better name I could have used?
2. The standard for configuration names seems to be `test#<configuration>#<variable_name>`. Which is why I used a separate config variable for basic and neovim strategies. Should I have just used `test#neovim#start_normal` for both `neovim` and `basic` strategies, since this only affects users runnin Neovim?

DanielPower:
Has anyone had a chance to look at this yet? @codeinabox? I'll be happy to make any required changes to get this in :)

codeinabox (approved):
Thank you for this and your patience, all seems sensible to me!

Issue #324, opened 2018-09-24
Testing vue-clie javascript components with jest: nearest does not work


doits:
To run jest in vue-cli apps, one cannot use `jest` directly (see [explanation](https://www.npmjs.com/package/@vue/cli-plugin-unit-jest)):

> Note that directly running jest will fail because the Babel preset requires hints to make your code work in Node.js, so you must run your tests with vue-cli-service test:unit.

To make this a little bit more complicated, I use `yarn` for my dependency management. This means to run tests I have to do:

```js
yarn test:unit [jest options]
```

I configured vim-test to do this:

```vimscript
let test#javascript#jest#executable = 'yarn test:unit'
```

But the problem is that the `jest` script passes `--` explicitly as a command line arg. For example the resulting command run is:

```
yarn test:unit --no-coverage -t '^ApplicationForm\.vue test descriptions$' -- tests/unit/components/ApplicationForm.spec.js
```

This unfortunately tests the whole file `tests/unit/components/ApplicationForm.spec.js`, not only the example I wanted to run. The reason is:

`yarn` outputs this warning:

> Warning From Yarn 1.0 onwards, scripts don't require "--" for options to be forwarded. In a future version, any explicit "--" will be forwarded as-is to the scripts.

It seems to ignore everything before the two dashes `--`, so in fact it runs:

```
yarn test:unit tests/unit/components/ApplicationForm.spec.js
```

When I remove the two dashes `--` manually, it works like expected (and `yarn` does not output a warning):

```
yarn test:unit --no-coverage -t '^ApplicationForm\.vue test descriptions$' tests/unit/components/ApplicationForm.spec.js
```

The general questions are:

- Do you want to detect a `vue-cli` app and `yarn` automatically in this case and do everything right automatically without configuration? A starting point could be to search for `@vue/cli-plugin-unit-jest` in `package.json` to detect the use of `jest` in combination with `vue-cli` and a `yarn.lock` for `yarn`
- If not, is adding a config option to the `jest` test runner to *not* add those two dashes `--` an option? (this would simply solve my problem at least)
- How can this scenario be handled in a neat way?

codeinabox:
[...]

PR #582, opened 2021-08-12
Use vue-cli-service to run tests

typedin:
When a project has been created with the vue-cli tool, it might a bit wiser to use the wrapper around the test runners that are installed via the command line tool.

maybe related to this: #324


spec/vue_test_utils.vim
```diff
@@ -0,0 +1,81 @@
...
+    call Teardown()
+    cd -
+  end
+
+  context "on nearest tests"
+    it "runs vue-cli-service"
+      view +1 __tests__/normal-test.js
+      TestNearest
+
+      " vue-cli-service test:unit --no-coverage -t \"renders props.msg when passed" -- tests/unit/example.spec.ts
```
codeinabox, line 19: Could you remove this comment

codeinabox (commented):
Thank you for this! Only two small pieces of feedback

1. Remove the comment
2. Update doc/test.txt with something like the following:

```
                                                *test-:VueTestUtils*
:VueTestUtils [args]  	         Uses the `vue-cli-service` command.
```

spec/vue_test_utils.vim
typedin: Sorry for that. It should be ok now.

PR #588, opened 2021-09-29
Add support for mint test

esensar:
This adds support for test runner for [mint language](https://www.mint-lang.com/)


codeinabox (approved):
Thank you so much for this! I didn't know about Mint so it's great to support another language

PR #598, opened 2021-10-31
Add TestNearest support for Lua's Busted test runner

mblayman:
This PR updates the `busted` plugin to add support for `TestNearest` so that it will execute the nearest test instead of the entire file (which is the current implementation). The modifications handle both `it` and `describe` functions. I also wrote some tests for Moonscript to make sure that it works for that style of Lua too.

I made this work by taking advantage of the `--filter` flag with `busted`. The `--filter` flag expects the value to be a proper Lua pattern. Because of that constraint, the plugin does the proper escaping of the Lua pattern "magic characters" (defined in the [Lua docs](https://www.lua.org/pil/20.2.html)).

There is a small limitation to this implementation. If multiple tests in the same file have the same string description, then `TestNearest` will run all identically labeled tests. This seemed like an acceptable tradeoff because it's still more accurate than running all the tests in a file.

PR Checklist:


autoload/test/base.vim
```diff
@@ -93,7 +93,7 @@ endfunction
 " Instead of taking a "position" argument, this function takes 3:
 "   - "filename" is the equivalent of "position['file']"
 "   - "from_line" the line number from where to start the search, is the
-"   equivalent fo "position['line']"
+"   equivalent of "position['line']"
```
codeinabox, line 96: Thank you for fixing this typo!

codeinabox (approved):
Thank you for taking the time for this great addition, it's super appreciated!

PR #601, opened 2021-11-10
Enable backtick surrounded test names in Kotlin

marcelja:
In Kotlin, test function names can be surrounded by backticks. https://todd.ginsberg.com/post/kotlin-function-names/

This PR would update the regex (to match any character between backticks) and replace the backticks in the gradle command by `"`.

README and docs updates not needed.


codeinabox (approved):
Perfect! Thank you for fixing this

PR #608, opened 2021-12-17
Use python -m pytest instead of pytest

### LINKED ISSUE #606: Change pytest executable to python -m pytest
When use `pytest` instead of `python -m pytest`, the `sys.path` may not be correctly set, cause the tests fail with `ImportError`. Here is a stackoverflow [question](https://stackoverflow.com/questions/42996270/change-pytest-rootdir) related to this.

There is a very simple fix, add `python -m` as the executable prefix when `pipenv` and `poetry` not in used. Basic add two lines in `autoload/test/python/pytest.vim`:

```vim
function! test#python#pytest#executable() abort
  let pipenv_prefix = ""

  if filereadable("Pipfile")
  elseif filereadable("poetry.lock")
    let pipenv_prefix = "poetry run "
  else
    let pipenv_prefix = "python -m"   " use python -m pytest instead of pytest
  endif

  if executable("py.test") && !executable("pytest")
    return pipenv_prefix . "py.test"
  else
    return pipenv_prefix . "pytest"
  endif
endfunction
```

I have tested this locally and it works fine. If this change is OK I can submit a pull request.

kkpattern:
When executing pytest command directly. The current directory won't be added to sys.path. In a typical python project structure, if we have a `package` directory and a `test` directory in the top directory, executing pytest directly will cause `from package import x` to fail in test code. This resolves #606 .


codeinabox (approved):
Thank you for this fix! Very much appreciated

PR #609, opened 2021-12-18
fix test function pattern for Go

lunarxlark:
In Go, test function name rule is to begin with `Test`.
Actually, `Testテスト` is testable.

```
> go test -v . --run Testテスト
=== RUN   Testテスト
--- PASS: Testテスト (0.00s)
PASS
ok      github.com/lunarxlark/test     0.243s
```

### DIFF
--- a/autoload/test/go.vim
+++ b/autoload/test/go.vim
@@ -1,5 +1,5 @@
 let test#go#patterns = {
-  \ 'test':      ['\v^\s*func (\w+)'],
+  \ 'test':      ['\v^\s*func ((Test|Example).*)\('],
   \ 'namespace': [],
 \}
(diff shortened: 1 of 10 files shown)


codeinabox (commented):
どうもありがとうございますよ! Thank you for this fix, should we also update the fixtures and specs to reflect this change?

lunarxlark:
> Thank you for this fix, should we also update the fixtures and specs to reflect this change?

Yes. I'm missed it. fix it later.

lunarxlark:
update fixture and spec of gotest, delve, richgo.

codeinabox (approved):
This is great and very thorough - thank you!

PR #614, opened 2021-12-29
[golang] Match methods as test functions

kassio:


codeinabox:
@kassio would you like a hand with this?

kassio:
@codeinabox, 

Yes, I'm currently without time, but I think would be nice to add some tests cases for this. What do you think?

codeinabox:
> @codeinabox,
> 
> Yes, I'm currently without time, but I think would be nice to add some tests cases for this. What do you think?

Thank you for patience on this, could give it a go

PR #615, opened 2021-12-30
Allow configuring stack test command

drewolson:
This PR allows overriding the test command for Haskell's `stack`. Specifically, it allows you to use `test .` as the default command, which works better in multi-package stack projects.

### DIFF
--- a/autoload/test/haskell/stacktest.vim
+++ b/autoload/test/haskell/stacktest.vim
@@ -2,6 +2,10 @@ if !exists('g:test#haskell#stacktest#file_pattern')
   let g:test#haskell#stacktest#file_pattern = '\v^(.*spec.*)\c\.hs$'
 endif
 
+if !exists('g:test#haskell#stacktest#test_command')
+  let g:test#haskell#stacktest#test_command = 'test'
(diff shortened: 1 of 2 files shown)


codeinabox (commented):
Thank you for this. There is a failing test in spec/stacktest_spec.vim, let me know if you need a hand updating this.

drewolson:
Whoops sorry about that, I'll take a look now.

drewolson:
I'm actually not sure why `"test"` was quoted only in the zero-argument case. I think this is incorrect. Are you ok with me updating this test to remove the quotes?

drewolson:
I've updated the test, let me know if this works for you.

codeinabox:
> I've updated the test, let me know if this works for you.

Thank you for quickly sorting this. Could you please  add a scenario to the spec where `g:test#haskell#stacktest#test_command` is overriden and then I think this is good to merge in.

drewolson:
@codeinabox done

codeinabox (approved):
Thanks again, this is great and I'm happy to merge it in

PR #635, opened 2022-03-10
fix(rust): support actix rt tests

rcarriga:
In case there are other prefixes, used an arbitrary word pattern rather
than hardcoded string.


codeinabox (approved):
Thank you for this fix, much appreciated!

PR #641, opened 2022-03-30
Support Go +build tags

Frederick888:

When current file's got build tag constraints, by default it may not be
compiled by `go test`. This commit tries to parse the +build tags [1]
and add `-tags=foo,bar` to the command line.

The new go:build [2] is not supported since it's quite troublesome to
process. Users of Go 1.17+ can have both go:build and +build in a file
as long as they are consistent, or manually `:TestFile -tags=foo,bar`.

[1] https://pkg.go.dev/cmd/go@go1.16.15#hdr-Build_constraints
[2] https://pkg.go.dev/cmd/go@go1.18#hdr-Build_constraints

### DIFF
--- a/autoload/test/go/gotest.vim
+++ b/autoload/test/go/gotest.vim
@@ -22,7 +22,31 @@ function! test#go#gotest#build_position(type, position) abort
 endfunction
 
 function! test#go#gotest#build_args(args) abort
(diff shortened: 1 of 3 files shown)


codeinabox (approved):
Thank you for this, much appreciated

PR #646, opened 2022-04-22
cargotest: Use cargo to determine package name

lluki:
Execute nearest fails when using a cargo workspace where the package name does not match the directory name fails. `cargotest.vim` uses the directory name, but a different package name can be specified in the `Cargo.toml`. 
This PR determines the package name using `cargo read-manifest`. 

A minimal example that is fixed by this PR:

Directory structure:
```
.
├── Cargo.toml
└── pack1
    ├── Cargo.toml
    └── src
        └── main.rs
```

`Cargo.toml`
```
[workspace]
members = [ "pack1", ]
```

`pack1/Cargo.toml`
```
[package]
name = "sim2"
version = "0.1.0"
edition = "2021"

# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html

[dependencies]
```

`pack1/src/main.rs`
```
fn main() {
    println!("Hello, world!");
}

// Execute nearest test on this line
#[test]
fn my_test() {}
```

### DIFF
--- a/autoload/test/rust/cargotest.vim
+++ b/autoload/test/rust/cargotest.vim
@@ -95,12 +95,20 @@ function! s:test_namespace(filename) abort
   endif
 
   let l:package = v:null
-  " Find package by searching upwards for Cargo.toml
+
+  " Find package by parsing cargo read-manifest output
+  let l:save_dir = chdir(fnamemodify(a:filename, ':p:h'))
+  if l:save_dir != ""
+    let l:json = system("cargo read-manifest")
+    let l:package = json_decode(l:json)['name']
+    call chdir(l:save_dir)
+  endif
+
+  " Cut modules to be relative to Cargo.toml
   for idx in range(len(l:modules) - 2, 0, -1)
       let l:cargo_toml = join(l:modules[:idx] + ['Cargo.toml'], '/')
       if !empty(glob(cargo_toml))
           echo 
-          let l:package = l:modules[idx]
           let l:modules = l:modules[idx+1:]
           break
       endif


lluki:
Hm. Is adding a dependency to `cargo` an option for you?

codeinabox:
> Hm. Is adding a dependency to `cargo` an option for you?

Thank you for this! May I suggest trying to install Cargo as part of the `before_script` inside `.travis.yml`. Let me know how you get on

nitnelave:
> I think I have a solution to testing this where we can mock the system call. Could you provide an example of what is returned by `cargo read-manifest`

`read-manifest` is deprecated. The modern version is something along these lines:
```sh
cargo metadata --no-deps --offline --format-version 1
```

And it returns something like:
```json
{
  "packages": [
    {
      "name": "sim2",
      "version": "0.1.0",
      "id": "sim2 0.1.0 (path+file:///project-root/pack1)",
      "manifest_path": "/project-root/pack1/Cargo.toml",
      "..."
    },
    "..."
  ],
  "workspace_members": [
    "sim2 0.1.0 (path+file:///project-root/pack1)",
    "..."
  ],
  "target_directory": "/project-root/target",
  "version": 1,
  "workspace_root": "/project-root"
}
```

Does that help?

codeinabox:
Not forgotten about this, will have a look into it soon

codeinabox:
I've made a start on trying to mock cargo with an alias though getting some errors, hopefully the answer will reveal itself tomorrow

spec/fixtures/cargo/crate/mock-manifest.json
```diff
@@ -0,0 +1,14 @@
...
+		"id": "sim2 0.1.0 (path+file:///project-root/pack1)",
+		"manifest_path": "/project-root/pack1/Cargo.toml"
+	}],
+	"workspace_members": [
+		"sim2 0.1.0 (path+file:///project-root/pack1)"
+	],
+	"target_directory": "/project-root/target",
+	"version": 1,
+	"workspace_root": "/project-root"
+}
```
codeinabox, line 14: This needs to be updated with the output of `cargo read-manifest`

codeinabox (changes requested):
Thank you for your patience on this, I've finally got the mock of `cargo read-manifest` working so that this should work on CI. However I've discovered that even with cargo installed, this PR is lacking a `Cargo.toml` in the fixture directory. Could you please add one and update the manifest mock accordingly?

PR #648, opened 2022-04-28
Distinguish test cases that don't start with "it" and have the same text

mkonig:
The following exmple shows the issue:
class BeanSpec extends AnyFlatSpec {
  "Decimal" should "throw error when initialized with non numeric string" in {
...
  }

  "Binary" should "throw error when initialized with non numeric string" in {
...
  }
}

When calling testnearest,
before the change both test cases are discovered and both are run at the same time no matter
where the cursor is placed.

In addition the changeset adds the possibility to 
set g:test#scala#pattern in your vim config file and overwrite the default.


codeinabox (approved):
Thank you for this improvement, much appreciated!

Issue #633, opened 2022-02-23
Cannot find test cases using nx runner 


perfectspr:
Nx cannot find test causes because it use the default project name. 
Below solution can automatically find the correct project for a test file. 

```vim
function! test#javascript#nx#build_position(type, position) abort
  let project = ''
  if filereadable('workspace.json')
    let l:workpaces = readfile('workspace.json')
    if exists('*json_decode')
      let l:projects = json_decode(join(workpaces, ''))['projects']
      for [key, value] in items(projects)
        if stridx(a:position['file'], value['root']) >= 0 
          let project = key
          break
        endif
      endfor
    endif
  endif

  if a:type ==# 'nearest'
    let name = s:nearest_test(a:position)
    if !empty(name)
      let name = '-t '.shellescape(name, 1)
    endif
    return [project, name, '--test-file', a:position['file'], '--no-coverage']
  elseif a:type ==# 'file'
    return [project, '--test-file', a:position['file'], '--no-coverage']
  else
    return [project, '--no-coverage']
  endif
endfunction
```

jponc:
Haven't tested the solution yet but yeah I'm having issues with NX + vim-test as well...

I'm getting

<img width="775" alt="image" src="https://user-images.githubusercontent.com/4714727/160743215-cf938ff4-ef84-4c03-b76a-c790e126e52c.png">

I'm not sure if it's related though..

phantomwhale:
That fixed it up for me! You should put that into a Pull Request :)

jponc:
Thanks for replying back @phantomwhale , do you happen to know what configuration I should do so it automatically detects it's an nx project instead of a usual JS?

excalios:
I had to tweak the code a bit to work on mine `value['root']` would show the first character only. So I had to change it to use `value` instead.

codeinabox:
Closing this as a solution has been merged in. Feel free to reopen if it doesn't resolve the issue

PR #649, opened 2022-04-28
(fix) nx project name

excalios:
NX  was using base project as project name. This PR changes the project name to the correct one.
Fixed by @perfectspr referenced in #633 

I don't think the README nor Vim documentation needs to be updated. For the spec I'm not sure on how to update it.


codeinabox:
> For the spec I'm not sure on how to update it.

Thank you for this addition and your patience for it to be reviewed. This looks great however I'd like to have a spec to deal with this scenario, happy to write this for you if you give me write access to your branch.

codeinabox:
Keen to get this merged but I want a spec for it. I've tried adding one but I am not sure on an example, so could you please help with this.

codeinabox:
Happy new year! Following this up as I'm keen to get it merged in, so let me know if you need a hand writing the spec

excalios:
Hello! sorry for the late reply. Yes!! help with writing the spec would be highly appreciated.

There is another problem though besides the nx wasn't able to find the correct project name. Now with the current nx version the structure of the nx project has changed to the point vim-test doesn't even detect the project anymore. Since now the project is only defined in the `project.json` in each project but there is no more in the root folder. Hence, making the test won't run at all since they can't find the project name. Currently, I have no solution for this problem though so far I just ran my test manually unfortunately.

excalios:
Hello, I'm back to this issue to mention I just tried to fix/adjust the test function for nx to adjust to the newest Nx structure which is using `project.json`. I'm still unsure about creating the spec file for the PR. I also didn't cover if we have a different nx version that is still using `workspace.json`. So, I'll just share the changes I made so other people that have the same experience as mine can adjust their code.

`autoload/test/javascript/nx.vim`

```vim
function! test#javascript#nx#build_position(type, position) abort
  let l:path = split(a:position['file'], '/')[0:-2]
  while len(path) > 0
    if filereadable(join(path, '/')."/project.json")
      let l:workspaces = readfile(join(path, '/')."/project.json")
      if exists('*json_decode')
        let l:name = json_decode(join(workspaces, ''))['name']
        let project = name
        break
      endif
    endif
    let path = path[0:-2]
  endwhile

  if a:type ==# 'nearest'
    let name = s:nearest_test(a:position)
    if !empty(name)
      let name = '-t '.shellescape(name, 1)
    endif
    return [project, name, '--test-file', a:position['file']]
  elseif a:type ==# 'file'
    return [project, '--test-file', a:position['file']]
  else
    return [project] " This will run the test for the project
  endif
endfunction
```

codeinabox (changes requested):
Thank you for getting back in touch, could you please update `spec/fixtures/nx/workspace.json` with an example that you developed against

excalios:
Sure! I'll try it in my spare time maybe over the weekend. I'll create a new PR later and I'll contact if I need any help

codeinabox:
> Sure! I'll try it in my spare time maybe over the weekend. I'll create a new PR later and I'll contact if I need any help

Thank you! No need to create a new PR, you could update this one

codeinabox:
Closing this as  #760, an implementation with fixtures, has been merged in

PR #656, opened 2022-05-26
Support go subtests

muchzill4:
Makes `:TestNearest` work with `t.Run()` subtests in golang.

Since `-run` takes a regex, there's a need to escape characters. I could use some guidance with tests for for those, as it is a bit nasty. 😅

---

- [ ] Update the README accordingly **no need?**
- [ ] Update the Vim documentation in `doc/test.txt` **no need?**


codeinabox (approved):
Thank you for this, much appreciated

PR #658, opened 2022-05-30
Stacktest improvements

MrcJkb:

- [ ] Update the README accordingly  (not applicable)
- [ ] Update the Vim documentation in `doc/test.txt` (not applicable)

## Improvements to the `stacktest` runner

* Add support for projects with subpackages
* Detect fully-qualified module name when running `TestFile`
* Add pattern for detection of hspec `prop` (property) tests
* Modify test patterns so that they work for qualified imports

Something I would like to add eventually, but don'.t have the time for right now:
(I have prepared test cases, which I have commented out for now)

* Support for detecting the spec tree, for example:
```haskell
describe "Given ..." $
  describe "And..." $
   it "Then..." $ do
    x `shouldBe` y
```
would result in
```
stack test --test-arguments "-m \"/Given.../And.../Then...\""
```

 I am not sure about the implementation of the module name detection, because it abuses the `get_nearest` function, and might fail if there is a comment before the module declaration that matches the search pattern.
 An alternative approach could be to  analyse the directory tree until a folder that starts with a lower case character is found.


codeinabox (commented):
Thank you for this and apologies it has taken me a while to review. Could you confirm if you want to leave in the TODO bits and I'll merge accordingly

MrcJkb:
> Thank you for this and apologies it has taken me a while to review. Could you confirm if you want to leave in the TODO bits and I'll merge accordingly

No pressure :)
Yes, it will probably be a while before I have time to try and implement the rest. It is currently in a stable state and can be merged. If you like, I can remove the TODO comments for now.

codeinabox (approved):
Thanks again for all your hard work on this!

MrcJkb:
Always glad to contribute to an awesome project :)

PR #664, opened 2022-06-30
Add option for harpoon strategy to not go to the terminal

matt-gardner:
When using the harpoon strategy, if the terminal is already open in a window, the current behavior will change the current window to also go to that terminal, which is not desirable.  This PR adds an option to disable this behavior, so that the user can do their own scripting to check whether to go to the terminal or not.

I haven't used vimscript much - I'm not sure if there's a better way than what I did to pass a local variable from vimscript to lua.

---

For the first two boxes, (1) this updates a strategy, not a runner, and I didn't see any relevant existing spec, and (2) it seemed too verbose to add this option to the harpoon row in the table in the README.  Let me know if I missed something.


codeinabox (approved):
Thank you for this simple and elegant solution! 🙏🏻

PR #675, opened 2022-08-03
Fix typo in variable name

camgraff:
Fixes a bug introduced in a5b122e8c5eb47db5a034908172ccbee44f4e520 where it will crash due to unknown variable


codeinabox (approved):
Thank you for the fix!

PR #683, opened 2022-08-22
Uses elm-test in node_modules if available

IzumiSy:
Elm runner does not cover the case that users install elm executables as their local dependecies.

This PR adds it.

---

### DIFF
--- a/spec/elmtest_spec.vim
+++ b/spec/elmtest_spec.vim
@@ -32,18 +32,36 @@ describe "elm-test"
   end
 
   it "runs tests against absolute path of npm executable (elm-test)"
-    let g:test#javascript#elm#executable = 'node_modules/.bin/elm-test'
-    view tests/NormalTest.elm
-    TestFile
+    try
+      !mkdir -p node_modules/.bin
+      !touch node_modules/.bin/elm-test
 
-    Expect g:test#last_command == 'node_modules/.bin/elm-test tests/NormalTest.elm'
+      view tests/NormalTest.elm
+      TestFile
+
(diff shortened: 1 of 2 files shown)


spec/elmtest_spec.vim
```diff
@@ -31,4 +31,19 @@ describe "elm-test"
     Expect g:test#last_command == 'elm-test'
   end
 
+  it "runs tests against absolute path of npm executable (elm-test)"
+    let g:test#javascript#elm#executable = 'node_modules/.bin/elm-test'
```
codeinabox, line 35: Instead of setting this variable could you test this by creating `node_modules/.bin/elm-test` using `touch` like in cucumberjs_spec.vim

spec/elmtest_spec.vim
```diff
@@ -31,4 +31,19 @@ describe "elm-test"
...
+    view tests/NormalTest.elm
+    TestFile
+
+    Expect g:test#last_command == 'node_modules/.bin/elm-test tests/NormalTest.elm'
+  end
+
+  it "runs tests against absolute path of npm executable (elm-test and elm-compiler)"
+    let g:test#javascript#elm#executable = 'node_modules/.bin/elm-test --compiler node_modules/.bin/elm'
```
codeinabox, line 43: Please use `touch` here too

codeinabox (changes requested):
Thank you for this, just some minor feedback about how to test this

codeinabox (approved):
Thank you for updating the tests, looks great

PR #688, opened 2022-08-31
Fix kitty strategy for Kitty 0.26.0

madalinoprea:
Since Kitty 0.26.0, vim-test fails to run the tests using `kitty`
strategy: a new Kitty window is created, but the command running the tests
is not executed.

The behavior is restored by replacing the usage of deprecated `new-window`
command with `launch`


codeinabox:
Thank you for this, only thought is should we have any feature / version detection in case someone is using an older version of Kitty or only support the latest version?

madalinoprea:
I wasn't able to figure out when `launch` command was added in Kitty and if it's backwards compatible. So I ended up following your suggestion to add version detection: use `launch` only for Kitty versions starting `0.26.0`.

codeinabox (approved):
On further reflection I am happy to merge this in as Kitty hasn't reached version 1 yet, thanks again!

PR #692, opened 2022-09-06
Add support for Nim's UnitTest

pusewicz:
Using standard library [https://nim-lang.org/docs/unittest.html](unittest)

Filtering happens via supplying test name as the last argument.

Running the whole `suite` happens via supplying the name with `::` suffix, which is the separator used between the suite name and a particular test name.


codeinabox (approved):
Thank you, love watching the list of languages that this runner supports grow!

PR #697, opened 2022-09-29
Support default cypress extension type

willfish:
I use cypress in my workplace and we've recently moved to the default naming
convention specified by some cypress config.

This PR moves vim-test to support this file naming structure.

- Support .cy.{js,jsx,coffee} file suffix type
- Adds fixtures for cy and test file suffix types
- Adds coverage for cy and test file types


willfish:
@janko I don't suppose you mind me merging this small change, do you?

janko:
@willfish Not at all 😉. Not sure why there is a review required, but hopefully you can bypass that.

willfish:
It actually says I'm not privileged enough to do a merge, sadly :) @janko

janko (approved):

willfish:
Thanks muchly @janko

janko:
@codeinabox I would maybe suggest removing this branch protection, because there are only two maintainers.

PR #699, opened 2022-10-31
Fixes Issues with Vitest Implementation

harrisoncramer:
Not super good with Vimscript but this seems to fix the TestNearest funcitonality for me for Vitest, relates to https://github.com/vim-test/vim-test/issues/698

My intent as the issue mentions is to remove the `$` and `^` from the Regexes here.

If someone else has thoughts on how to fix this MR please let me know. Thanks.


spec/vitest_spec.vim
```diff
@@ -18,7 +18,23 @@ describe "vitest"
...
+      TestNearest
+
+      Expect g:test#last_command == g:expectedExecutable .. 'vitest run --no-coverage -t ''^Math Addition'' __tests__/normal-test.js'
+
+      view +3 __tests__/normal-test.js
+      TestNearest
+
+      Expect g:test#last_command == g:expectedExecutable .. 'vitest run --no-coverage -t ''^Math Addition adds two numbers$'' __tests__/normal-test.js'
+    end
+    it "runs JSX"
```
harrisoncramer, line 37: Vitest supports Javsacript out of the box and the default configuration should also support JS. The previous author had mislabeled JSX extensions as "Javascript" and TSX extensions as "React" which is a bit odd

maorun:
if we removes $ or ^ other tests will also run.
perhaps something is missing to use `-t` (testNamePattern)

```
describe('dont will', () => {
    it('should do good stuff', () => {
        console.log('a')
        expect(true).toBeTruthy()
    })
})

describe('will', () => {
    it('should do good stuff', () => {
        expect(true).toBeTruthy()
    })
    it('should do good stuff and other things', () => {
        expect(true).toBeTruthy()
    })
    it('should do good stuff but only on something', () => {
        expect(true).toBeTruthy()
    })
})
```

codeinabox:
Would you like a hand fixing these failing tests?

v-ostark:
Any update on this?

harrisoncramer:
Hey @codeinabox yes, a hand fixing the failing tests would be great!

codeinabox:
> Hey @codeinabox yes, a hand fixing the failing tests would be great!

I'll have a look into this next week

v-ostark:
Can we merge this?

codeinabox:
> Can we merge this?

Once we have resolved the tests failing on CI

codeinabox (changes requested):
The following test are failing:

### vitest runs file test if nearest test couldn't be found

Expected g:test#last_command == g:expectedExecutable .. 'vitest run --no-coverage __tests__/normal-test.jsx' at line 5

Actual value: "npx vitest run --no-coverage -t '' __tests__/normal-test.jsx"
Expected value: "npx vitest run --no-coverage __tests__/normal-test.jsx"

### vitest on nearest tests runs Javascript

Expected g:test#last_command == g:expectedExecutable .. 'vitest run --no-coverage -t ''Math'' __tests__/normal-test.js' at line 4

Actual value: "npx vitest run --no-coverage -t '' __tests__/normal-test.js"
Expected value: "npx vitest run --no-coverage -t 'Math' __tests__/normal-test.js"

codeinabox:
So I've managed to get the tests passing. Firstly I reverted the logic to only set `-t` when there is a name. I noticed the tests for normal-test.js you added were failing though it's not clear why, however removing them the tests are now passing. If there are no objects I am keen to get this merged in

codeinabox (approved):

PR #703, opened 2022-12-06
Pass files as path to jest

linas-deliveroo:
By default Jest treats file paths as regex patterns. As a consequence it fails to pick tests from Next.js files like `src/pages/api/post/[postId]/index.test.ts`. Pass a flag to make Jest treat those parameters as file paths rather than regexes. This is easier and cleaner than escaping special regex characters.

- [ ] Update the README accordingly (no update needed)
- [ ] Update the Vim documentation in `doc/test.txt` (no update needed)


codeinabox (approved):
I didn't know about this handy Jest feature - thank you!

PR #705, opened 2022-12-13
Add teenytest

pingortle:
This bakes my simple custom runner for TestDouble's [teenytest](https://github.com/testdouble/teenytest) package into vim-test. I followed the basic pattern of existing Ava integration as a template.

This is my first contribution here, and first vimscript PR ever, so I'd appreciate feedback on both form and function.

---


pingortle:
Hmmm… The `:Teenytest` command does not seem to exist with `vim-test` installed from my fork. I was under the impression that it would exist magically since I couldn't find definitions for other commands. I don't personally need that command and only added it to the docs for consistency. I would also happily revert the entry from `doc/test.vim`.

Can anyone provide guidance on this?

codeinabox (approved):
Thank you for this, looks great!

codeinabox:
> Hmmm… The `:Teenytest` command does not seem to exist with `vim-test` installed from my fork. I was under the impression that it would exist magically since I couldn't find definitions for other commands. I don't personally need that command and only added it to the docs for consistency. I would also happily revert the entry from `doc/test.vim`.
> 
> Can anyone provide guidance on this?

For the command to be available, you need to [add it to the runner commands configuration](https://github.com/vim-test/vim-test#runner-commands)

spec/fixtures/teenytest/test/normal.js
```diff
@@ -0,0 +1,5 @@
+module.exports = {
+  'Adds two numbers': function () {
+    // assertions
+  }
+}
```
codeinabox, line 5: Could you update this with an example from their docs, this is probably a good one to go with: https://github.com/testdouble/teenytest/blob/main/example/simple-node/test/lib/dog-test.js

codeinabox (changes requested):

spec/fixtures/teenytest/test/normal.js
pingortle: For sure! Something like this https://github.com/vim-test/vim-test/pull/705/commits/c329ec5570ade6810175b3976dedfe97b7f3f115?

codeinabox (approved):

Issue #708, opened 2023-01-29
TestNearest doesn't work with nested Rust test modules


dyllandry:
This might be by design, but when organizing rust tests using nested modules, TestNearest doesn't work.

```rust
#[cfg(test)]
mod tests {
    mod my_module {
        #[test]
        fn fail() {
            assert!(false);
        }
    }
}
```

With the cursor inside `fail` and running `:TestNearest`, I get:

```
    Finished test [unoptimized + debuginfo] target(s) in 0.20s
     Running unittests src/main.rs (target/debug/deps/rust_tic_tac_toe-39273d94b1802694)

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 1 filtered out; finished in 0.00s
```

dyllandry:
It works though after installing nextest.

dyllandry:
I found out it works if
1) nextest is installed
2) the module has the word "tests" in it

```rust
#[cfg(test)]
mod tests {
    mod my_module_tests {
        #[test]
        fn fail() {
            assert!(false);
        }
    }
}
```

With the cursor inside `fail` and running `:TestNearest`, I get:

```
    Finished test [unoptimized + debuginfo] target(s) in 0.00s
    Starting 1 tests across 1 binaries
        FAIL [   0.001s] rust-tic-tac-toe::bin/rust-tic-tac-toe tests::my_module_tests::fail

--- STDOUT:              rust-tic-tac-toe::bin/rust-tic-tac-toe tests::my_module_tests::fail ---

running 1 test
test tests::my_module_tests::fail ... FAILED

failures:

failures:
    tests::my_module_tests::fail

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s


--- STDERR:              rust-tic-tac-toe::bin/rust-tic-tac-toe tests::my_module_tests::fail ---
thread 'tests::my_module_tests::fail' panicked at 'assertion failed: false', src/main.rs:122:13
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

   Canceling due to test failure: 0 tests still running
------------
     Summary [   0.002s] 1 tests run: 0 passed, 1 failed, 0 skipped
        FAIL [   0.001s] rust-tic-tac-toe::bin/rust-tic-tac-toe tests::my_module_tests::fail
error: test run failed
```

dyllandry:
[...]

PR #713, opened 2023-02-12
add spec and fixture for rust nested test modules

dyllandry:

  - _Didn't, I don't think necessary._
  - _Didn't, I don't think necessary._

The added failing spec demonstrates the issue https://github.com/vim-test/vim-test/issues/708 

vim-test isn't running the correct test command for rust files, with both cargo test and nextest, when tests in a single file are nested inside modules. The linked issue has an example.

### DIFF
--- a/spec/cargonextest_spec.vim
+++ b/spec/cargonextest_spec.vim
@@ -115,6 +115,20 @@ describe "CargoNextest"
     Expect g:test#last_command == 'cargo nextest run ''too::nested::tests::second_test'''
   end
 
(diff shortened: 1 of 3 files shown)


dyllandry:
Some context from #708, I don't know much vimscript, but it was suggested I open a PR that adds a spec and fixture to demonstrate the problem. I _managed_ to do that. Let me know if I can improve it.

Hopefully someone else who knows more about vim-test than I can fix the underlying issue. If no one comes by for long enough that I've learnt enough vimscript, I'll try fixing it.

PR #719, opened 2023-02-22
Python: Use explicit Python3 in invocation

mihaigalos:
This is necessary since some platforms such as MacOS have no explicit alias for python.

### DIFF
--- a/autoload/test/python/pyunit.vim
+++ b/autoload/test/python/pyunit.vim
@@ -39,7 +39,11 @@ function! test#python#pyunit#executable() abort
     let pipenv_prefix = "pipenv run "
   endif
 
-  return pipenv_prefix . 'python3 -m unittest'
+  if strlen(system('command -v python3'))
+    return pipenv_prefix . 'python3 -m unittest'
+  else
+    return pipenv_prefix . 'python -m unittest'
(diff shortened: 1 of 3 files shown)


autoload/test/python/pyunit.vim
```diff
@@ -39,7 +39,7 @@ function! test#python#pyunit#executable() abort
     let pipenv_prefix = "pipenv run "
   endif
 
-  return pipenv_prefix . 'python -m unittest'
+  return pipenv_prefix . 'python3 -m unittest'
```
codeinabox, line 42: Instead of hard coding this use detection, for example

```viml
if strlen(system('command -v python3'))
  return pipenv_prefix . 'python3 -m unittest'
else
  return pipenv_prefix . 'python -m unittest'
endif
```

codeinabox (changes requested):

codeinabox (changes requested):
This is looking great, only thing remaining is to update `pyunit_pipenv_spec.vim` and `pyunit_spec.vim` to expect python3 as those tests are now failing on CI

codeinabox (approved):
Thank you for all your work on this

PR #720, opened 2023-03-05
phpunit: handle attributes added in phpunit 10

camilledejoye:
I left a commented test because I was not able to make it work reliably. As the fixtures shows it would mean matching a line with only `Test,` on it and that could be any comment or string on multiple lines as well.
Since this is an optional syntax and we can add as many attributes as possible I decided to not take the chance.


codeinabox (approved):
Thanks for this, great work!

As for multi-line matching, that's not supported as I came up against the same issue with #508

PR #733, opened 2023-04-28
bugfix(csharp) - fixing default slash for windows

joro550:

When running one of these commands for csharp in windows it was complaining that s:slash variable was not set - I think it might be because I am on windows and not using a terminal that equates to either cmd.exe or powershell, in this case the pack was not working so I have set the slash to be a forward slash unless you are running a specific terminal on windows

Fixes - https://github.com/vim-test/vim-test/issues/731


codeinabox (approved):
Thank you!

PR #735, opened 2023-05-19
Only use Sail when it's configured

jessarcher:
The current behaviour in vim-test is to use Laravel Sail whenever `vendor/bin/sail` exists, however, this comes pre-installed in fresh Laravel applications these days and does not mean the user will actually be using Docker for local development.

Currently users will get the following error when using vim-test in a fresh Laravel application:
image

Users that wish to use Sail would typically run `php artisan sail:install` which will create their `docker-compose.yml` file.

This PR updates vim-test to only use Sail when `docker-compose.yml` (or `docker-compose.yaml`) also exists in addition to `vendor/bin/sail`.


codeinabox (approved):
Thank you! This is a great solution

PR #736, opened 2023-06-23
Add Zig Support

bobcats:

I'm new to both vimscript and zig, so I mostly copied this from the `bats` implementation. From what I can gather, [there isn't a way to use a pattern to run a single test with `zig test`, so there's a chance multiple tests match](https://github.com/ziglang/zig/issues/13524), but this seems close enough for now?

For running the whole suite, seems like there's a pattern to define a test step in `build.zig`, [(1)](https://github.com/mattnite/gyro), [(2)](https://github.com/kooparse/zalgebra), so that's how I decided to implement `TestSuite`.

Either way, thanks for considering! Let me know if I can make any changes.

### DIFF
--- a/autoload/test/zig/zigtest.vim
+++ b/autoload/test/zig/zigtest.vim
@@ -2,30 +2,15 @@ if !exists('g:test#zig#zigtest#file_pattern')
   let g:test#zig#zigtest#file_pattern = '\v\.zig$'
 endif
 
-if !exists('g:test#zig#zigtest#test_patterns')
-  let g:test#zig#zigtest#test_patterns = {
-        \ 'test': [
-        \ '\vtest "=([^"]*)"= \{',
-        \ ],
-        \ 'namespace': []
-        \ }
-endif
-
-" Returns true if the given file belongs to your test runner
 function! test#zig#zigtest#test_file(file) abort
   return a:file =~# g:test#zig#zigtest#file_pattern
 endfunction
 
-" Returns test runner's arguments which will run the current file and/or line
 function! test#zig#zigtest#build_position(type, position) abort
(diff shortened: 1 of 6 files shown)


autoload/test/zig/zigtest.vim
```diff
@@ -0,0 +1,48 @@
...
+
+" Returns test runner's arguments which will run the current file and/or line
+function! test#zig#zigtest#build_position(type, position) abort
+  if a:type ==# 'nearest'
+    let name = test#base#nearest_test(a:position, g:test#zig#zigtest#test_patterns)
+
+    if empty(name['test'])
+      return []
+    else
+      return ['test', a:position['file'], '--test-filter', ''''.name['test'][0].'''']
```
codeinabox, line 27: > I'm new to both vimscript and zig, so I mostly copied this from the `bats` implementation. From what I can gather, [there isn't a way to use a pattern to run a single test with `zig test`, so there's a chance multiple tests match](https://github.com/ziglang/zig/issues/13524), but this seems close enough for now?

If that's the case, I'd suggest changing nearest to behave like a life

codeinabox, line 27: > I'm new to both vimscript and zig, so I mostly copied this from the `bats` implementation. From what I can gather, [there isn't a way to use a pattern to run a single test with `zig test`, so there's a chance multiple tests match](https://github.com/ziglang/zig/issues/13524), but this seems close enough for now?

If that's the case, I'd suggest changing nearest to behave like testing a file, and comment the code accordingly

```vim
if a:type ==# 'nearest'
    return ['test', a:position['file']]
```

codeinabox (approved):
Great work, thank you for this addition ❤️

PR #737, opened 2023-06-29
Add leintest for Clojure

dradtke:

---

This PR adds support for running Clojure tests via Leiningen, without requiring Fireplace.vim.


autoload/test/clojure/fireplacetest.vim
```diff
@@ -3,6 +3,11 @@ if !exists('g:test#clojure#fireplacetest#file_pattern')
 end
 
 function! test#clojure#fireplacetest#test_file(file) abort
+  if exists('g:test#clojure#runner') && g:test#clojure#runner ==# 'fireplacetest'
```
codeinabox, line 6: Is there a way to do this with zero configuration?

codeinabox: Is checking against `project.clj` a possibility? Will it contain any lines that indicate the project is using Lein?

dradtke: That may work, but the `fireplacetest` runner I think would also need to check for the presence of Fireplace.vim. My assumption here is that `fireplacetest` should be preferred for projects where Fireplace.vim is present, even if it's a Leiningen project.

I'll take a look at updating this to make both Clojure runners a little smarter there, but I think this `g:test#clojure#runner` check should remain for parity with other languages.

codeinabox: If other runners also use `g:test#clojure#runner` go with this approach now. Going forward, I'd like runners to adhere to a zero configuration required approach where detection is possible

codeinabox (approved):

autoload/test/clojure/fireplacetest.vim
dradtke: IMO it should be both. Auto-detection should pick a sane default, but these settings should allow that to be overridden if they are incorrect for whatever reason.

I'll push some simple autodetect logic to make sure it picks the right one for cases where Fireplace is not installed, and `project.clj` is present.

codeinabox: Thank you! That is how the JavaScript runner detection works too, it will check against the dependencies, or you can override it with configuration

codeinabox: Have you pushed all your changes, or should I hold off reviewing again and merging?

dradtke: All my changes have been pushed, thanks!

codeinabox (approved):

Issue #296, opened 2018-04-13
Support java integration tests (maven failsafe)


languitar:
With the maven failsafe plugin, integration tests are possible in case the tests follow a certain file name patterns:

```
<includes>
    <include>**/IT*.java</include>
    <include>**/*IT.java</include>
    <include>**/*ITCase.java</include>
</includes>
```
https://maven.apache.org/surefire/maven-failsafe-plugin/integration-test-mojo.html

Would be nice if this pattern was supported in the default configuration.

codeinabox:
Good suggestion, please feel free to raise a PR. Otherwise if you are able to add a scenario to the [Maven test spec](https://github.com/janko-m/vim-test/blob/master/spec/maventest_spec.vim), even if it's in pseudo code, that would help confirm the desired behaviour.

Tacsiazuma:
@languitar currently vim-test uses surefire-plugin to execute tests. I also need this behaviour and started to make some changes locally but wanna make sure we have the same goal.
So lets say you are in a buffer named `name.package.SomeTest.java`. Currently it will invoke: 

`mvn test -Dtest=name.package.SomeTest`

If we want this run by failsafe (this is what I've achieved so far) we need to do something like this:

`mvn verify -Dit.test=SomeTest`

But this needs a new command which I named as :IntegrationTest for now. Until other languages implement this kind of behaviour it would require manual configuration by adding something like this to your .vimrc:

`command! -nargs=* -bar IntegrationTest   call test#run('integration', split(<q-args>))`

If you want to run a pattern instead it can be done as well but that also needs a new command which accepts a pattern. Would this help? @codeinabox any thoughts?

languitar:
> But this needs a new command which I named as :IntegrationTest for now. Until other languages implement this kind of behaviour it would require manual configuration by adding something like this to your .vimrc:
> 
> `command! -nargs=* -bar IntegrationTest call test#run('integration', split(<q-args>))`

Another option could be to define a cascade of commands to execute, first unit and then integration tests. I think the original reason for my request was specifically to support `TestNearest`. That would need some dispatching then, which of the commands is relevant here.

GiuseppeMP:
@codeinabox Hey

I have implemented TestNearest and TestFile with failsafe here: [fix-#296](https://github.com/GiuseppeMP/vim-test/commit/f5d650d378024b2757929183b0dc61ad04f73a44)

[...]

PR #740, opened 2023-07-10
fix(failsafe): Support java integration tests (maven failsafe) #296 

GiuseppeMP:


GiuseppeMP:
@codeinabox It's working OK for me.
A few observations about performance, due the phase verify always runs unit tests and other tests like PIT, it results in a longer waiting time, because it depends on how many units tests and other tests that project have. 

In summary, the longer the waiting time for the unit suites and other tests, the longer the waiting time for `Nearest` and `TestFile` to finish.

I can write a recipe, how to run able mvn to run only the failsafe test and ignore the rest, but it needs to be configured in the pom.xml. 

Running isolated integration test is very cool with `:TestNearest`, I not sure, but it's not a present feature in others IDEs that I know.

codeinabox (approved):
Looks great, thank you for this!

PR #741, opened 2023-07-14
gotest: support verbose test

lotusirous:

### DIFF
--- a/autoload/test/go/gotest.vim
+++ b/autoload/test/go/gotest.vim
@@ -50,6 +50,11 @@ function! test#go#gotest#build_args(args) abort
 endfunction
 
 function! test#go#gotest#executable() abort
+  " Get the value of the global option g:test#go#gotest#test_verbose, default to 0 (false) if not set
+  let test_verbose = get(g:, 'test#go#gotest#test_verbose', 0)
+  if test_verbose
+    return 'go test -v'
+  endif
   return 'go test'
 endfunction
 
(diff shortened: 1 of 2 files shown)


doc/test.txt
```diff
@@ -626,6 +626,7 @@ You may find yourself specifying certain options for your test runners in
 certain situations. You can configure your preferred options with
 >
   let g:test#ruby#minitest#options = '--verbose'
+  let g:test#go#gotest#test_verbose = 1
```
codeinabox, line 629: Instead of introducing a new variable for something that vim-test already offers, I'd rather we add an example of in the docs.

```suggestion
  let g:test#go#gotest#options = '-v'
```

autoload/test/go/gotest.vim
```diff
@@ -50,6 +50,11 @@ function! test#go#gotest#build_args(args) abort
 endfunction
 
 function! test#go#gotest#executable() abort
+  " Get the value of the global option g:test#go#gotest#test_verbose, default to 0 (false) if not set
+  let test_verbose = get(g:, 'test#go#gotest#test_verbose', 0)
+  if test_verbose
+    return 'go test -v'
+  endif
```
codeinabox, line 57: This code is not necessary, vim-test already supports passing options

codeinabox (changes requested):
Thank you for your PR.  It's not necessary to introduce a new variable just for the verbose option, but worth updating the documentation with another example

doc/test.txt
lotusirous: Thank you for your suggestion

PR #754, opened 2023-08-15
zig: use --test-filter when running nearest test

mrnugget:
This changes the 'nearest' test type to use `zig test --test-filter`. It kinda goes against the decision made by @bobcats  in https://github.com/vim-test/vim-test/pull/736, which specifically didn't support this because `--test-filter` doesn't allow for regexs and might run more tests than we want.

But I figured: I'd rather run more tests if they match the same name, vs. the whole file all the time.

At least I want this feature, but happy to hear any feedback :)

---

### DIFF
--- a/autoload/test/zig/zigtest.vim
+++ b/autoload/test/zig/zigtest.vim
@@ -9,12 +9,11 @@ endfunction
 function! test#zig#zigtest#build_position(type, position) abort
   if a:type ==# 'nearest'
     let name = s:nearest_test(a:position)
-    if !empty(name)
-      let filter = '--test-filter '.shellescape(name, 1)
-      return ['test', a:position['file'], filter]
+    if empty(name)
+      return ['test', a:position['file']]
     endif
 
-    return ['test', a:position['file']]
+    return ['test', a:position['file'],  '--test-filter '.shellescape(name, 1)]
   elseif a:type ==# 'file'
     return ['test', a:position['file']]
   else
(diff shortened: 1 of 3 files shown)


codeinabox:
Quick question, if you specify both the test filter and a file, eg `zig test normal.zig --test-filter 'numbers 2'`, will that only run matching tests within that file?

autoload/test/zig/zigtest.vim
```diff
@@ -8,8 +8,12 @@ endfunction
...
   if a:type ==# 'nearest'
-    " --test-filter currently only supports filtering string inclusion, so
-    "  multiple tests could match, so instead we run the whole file.
+    let name = s:nearest_test(a:position)
+    if !empty(name)
+      let filter = '--test-filter '.shellescape(name, 1)
+      return ['test', a:position['file'], filter]
+    endif
+
     return ['test', a:position['file']]
```
codeinabox, line 17: This could be refactored to:

```suggestion
    if empty(name)
      return ['test', a:position['file']]
    endif

    return ['test', a:position['file'],  '--test-filter '.shellescape(name, 1)]
```

mrnugget:
> Quick question, if you specify both the test filter and a file, eg `zig test normal.zig --test-filter 'numbers 2'`, will that only run matching tests within that file?

Yep, exactly:

```
$ zig test numbers.zig
All 3 tests passed.
$ zig test numbers.zig --test-filter 'numbers'
All 2 tests passed.
$ zig test numbers.zig --test-filter 'numbers 2'
All 1 tests passed.
```

codeinabox (approved):
Thank you for this!

PR #762, opened 2023-09-27
Add Haskell test runner: cabaltest

niclas-ahden:
`vim-test` already has a runner for `stack` and this adds `cabal` as another option. Both `stack` and `cabal` are used throughout the Haskell ecosystem, so it makes sense to support them both.

This runner is based on the `stacktest` runner. I duplicated the fixtures etc. so that each runner is self-contained (I saw that that was the approach taken for other languages with multiple runners e.g. Rust).

`stacktest` remains as the default for Haskell so that we don't break the experience for current users.

Thank you for creating and maintaining `vim-test`, it's awesome!

## PR template

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ runners are supported:
 |     **Erlang** | CommonTest, EUnit, PropEr                                                                                          | `commontest`, `eunit`, `proper`                                                                                                              |
 |         **Go** | Ginkgo, Go, Rich-Go, Delve                                                                                         | `ginkgo`, `gotest`, `richgo`, `delve`                                                                                                        |
 |     **Groovy** | Maven, Gradle                                                                                                      | `maventest`, `gradletest`                                                                                                                    |
-|    **Haskell** | stack                                                                                                              | `stacktest`                                                                                                                                  |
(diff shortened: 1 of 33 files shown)


niclas-ahden:
Tests are passing and this is ready for review 👌

codeinabox:
Thank you for this and the kind words. Do both runners use Hspec for the tests and the only difference is the executable you use to run the tests, or is it more nuanced than that?

niclas-ahden:
@codeinabox My pleasure!

Yes, both use HSpec, and the differences are:

- The test args are different
- The project and package files in the fixtures, and the matchers for finding them, are different (and conflicting)
- They handle escaping of test names differently

This could surely be abstracted, but I didn't invest the time. I usually abstract when I get to three occurrences and here we're at two. I don't foresee alternatives to `stack` and `cabal` popping up, but we may want to support more test frameworks in the future (beyond HSpec). If we do, we'd have to abstract at that point as we'd have to manage the matrix of package managers vs. test frameworks.

codeinabox (approved):
Thank you for this and the detailed explanation!

Issue #763, opened 2023-10-08
vimterminal strategy does not unload buffer


chrisortman:
When I run tests (ruby / rake) using the vimterminal strategy in vim 9 (both brew vim and macvim) the terminal window opens and runs the test. Pressing <enter> will close the terminal window but the terminal window remains in my buffer list. 

Every test run creates a new buffer in the list and does not remove them.

This issue https://github.com/vim/vim/issues/5768 is maybe a little similar or had some useful suggestions, didn't solve the problem.

Using `:bd` instead of `:q` in the strategy function does seem to work.

PR #764, opened 2023-10-18
fix: delete buffer when closing vimterminal window

hugovandevliert:
As mentioned in #763, pressing `<Enter>` in a vim terminal window closes it, but does not unload the buffer. This causes every test run to leave a lingering buffer behind, polluting the buffer list.

As far as I can tell, using `:bd` instead of `:q` to make sure the buffer is deleted does not have any drawbacks.

### DIFF
--- a/autoload/test/strategy.vim
+++ b/autoload/test/strategy.vim
@@ -88,7 +88,7 @@ function! test#strategy#vimterminal(cmd) abort
   execute term_position . ' new'
   call term_start(!s:Windows() ? ['/bin/sh', '-c', a:cmd] : ['cmd.exe', '/c', a:cmd], {'curwin': 1, 'term_name': a:cmd})
   au BufLeave <buffer> wincmd p
-  nnoremap <buffer> <Enter> :q<CR>
+  nnoremap <buffer> <Enter> :bd<CR>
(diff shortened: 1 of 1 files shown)


autoload/test/strategy.vim
```diff
@@ -87,8 +87,8 @@ function! test#strategy#vimterminal(cmd) abort
   let term_position = get(g:, 'test#vim#term_position', 'botright')
   execute term_position . ' new'
   call term_start(!s:Windows() ? ['/bin/sh', '-c', a:cmd] : ['cmd.exe', '/c', a:cmd], {'curwin': 1, 'term_name': a:cmd})
-  au BufLeave <buffer> wincmd p
-  nnoremap <buffer> <Enter> :q<CR>
+  au BufDelete <buffer> wincmd p " switch back to last window
```
hugovandevliert, line 90: Bonus fix: we can now use `BufDelete` instead (like we already do for the `neovim` strategy) since we're actually deleting buffers. This was the reason `BufDelete` didn't work in the first place, as mentioned in #341.

codeinabox (approved):
Thank you for the fix, much appreciated!

PR #774, opened 2023-12-14
clojure: Expand patterns to support more use-cases

dradtke:

This PR updates the Clojure test patterns to support a few additional use-cases. The pattern originally only supported tests defined as `(deftest ...`, but with this change now also supports

1. Qualified `deftest` (i.e. `(clojure.test/deftest ...`)
2. [`defspec`](http://clojure.github.io/test.check/clojure.test.check.clojure-test.html#var-defspec)
3. Qualified `defspec`

It also updates the Leiningen test runner to wrap the test name in `shellescape()`, since Clojure supports a number of additional characters that wreak havoc with Bash.


codeinabox (approved):
Thank you for this great addition!

PR #694, opened 2022-09-13
Added test suite method pattern

mats852:
Test suites such as stretchr/testify uses methods as test functions

Make sure these boxes are checked before submitting your pull request:

- [x] Add fixtures and spec when implementing or updating a test runner
- [x] Update the README accordingly
- [x] Update the Vim documentation in `doc/test.txt`

PR #781, opened 2024-01-06
feat: support for go stretchr/testify

FreedomKnight:

Hi, I  implement the issue https://github.com/vim-test/vim-test/issues/624
and I think the document  is added by PR #694
I hope I can help it.

And It supposes that the user imports "testify/suite" and there is only one general Go test (the test has testing.T) as Test{X}Suite.

### DIFF
--- a/autoload/test/go/gotest.vim
+++ b/autoload/test/go/gotest.vim
@@ -15,6 +15,12 @@ function! test#go#gotest#build_position(type, position) abort
     if a:type ==# 'file'
       return path ==# './.' ? [] : [path . '/...']
     elseif a:type ==# 'nearest'
+      let matched_testify = s:is_testify()
+      if matched_testify == 1
+          let suite_name = s:get_suite_name()
+          let name = s:nearest_test(a:position)
+          return empty(name) ? [] : [path, '-run '.shellescape(suite_name.'$') . ' -testify.m ' .shellescape(name, 1)]
+      endif
       let name = s:nearest_test(a:position)
       return empty(name) ? [] : ['-run '.shellescape(name.'$', 1), path]
     endif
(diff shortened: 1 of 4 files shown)


codeinabox (changes requested):
Thank you for taking the time to make this PR! Looking great so far, just couple of outstanding things

-  Update README and doc/test.txt with details about `let g:test#go#gotest#testify#enabled = 1`
- Update `spec/gotest_spec.vim` with an example where the argument  `-testify.m`  is expected, let me know if you need a hand with that

autoload/test/go/gotest.vim
```diff
@@ -57,10 +67,32 @@ function! test#go#gotest#executable() abort
...
+  endfor
+  return ""
+endfunction
+
 function! s:nearest_test(position) abort
   let name = test#base#nearest_test(a:position, g:test#go#patterns)
   let name = join(name['namespace'] + name['test'], '/')
   let without_spaces = substitute(name, '\s', '_', 'g')
   let escaped_regex = substitute(without_spaces, '\([\[\].*+?|$^()]\)', '\\\1', 'g')
+  echo escaped_regex
```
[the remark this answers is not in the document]
FreedomKnight: Done

autoload/test/go/gotest.vim
```diff
@@ -15,6 +19,12 @@ function! test#go#gotest#build_position(type, position) abort
     if a:type ==# 'file'
       return path ==# './.' ? [] : [path . '/...']
     elseif a:type ==# 'nearest'
+       
+      if g:test#go#gotest#testify#enabled == 1 && s:is_testify() == 1
```
codeinabox, line 23: Is the flag `g:test#go#gotest#testify#enabled` needed when there is the `s:is_testify()` check? It appears it could be removed, making this change require zero configuration, which is the ethos of this plugin

FreedomKnight: I think it should be kept because some developers may mix two way in the same codebase

codeinabox: As long as removing it doesn't break any existing tests, I'd rather we remove the need for any additional configuration and only introduce it if and when a need arises. It also keeps the PR simpler as there is no need to update any documentation

FreedomKnight:
Thank you for your feedback.
I've updated the readme.md  in the Language-specific section.
However, I am uncertain about whether I should also include this update in doc/test.txt, as I didn't find a section there that specifically addresses language-specific configuration

autoload/test/go/gotest.vim
FreedomKnight: OK, I've removed the flag to keep it simple

codeinabox:
> Thank you for your feedback.
> I've updated the readme.md  in the Language-specific section.
> However, I am uncertain about whether I should also include this update in doc/test.txt, as I didn't find a section there that specifically addresses language-specific configuration

As the flag has been removed there is no need to update the documentation

codeinabox (approved):
Thank you, this looks great!

PR #783, opened 2024-01-08
refactor: pyunit logic to check for import unittest

codeinabox:


amosalb:
Hello! Today, I updated the vim-test and it was not more possible run my python tests. Until the update it worked well. Every time that I ran :Test* returns "Not a test file".
It seems that the refactoring broken something. Well, I set my vim-test to a previous commit, two weeks ago, and worked well.

codeinabox:
> Hello! Today, I updated the vim-test and it was not more possible run my python tests. Until the update it worked well. Every time that I ran :Test* returns "Not a test file".
> It seems that the refactoring broken something. Well, I set my vim-test to a previous commit, two weeks ago, and worked well.

@amosalb Thank you for reporting this, could you please give an example of the test file name and which Python test framework you expected it to use

codeinabox:
> Hello! Today, I updated the vim-test and it was not more possible run my python tests. Until the update it worked well. Every time that I ran :Test* returns "Not a test file".
> It seems that the refactoring broken something. Well, I set my vim-test to a previous commit, two weeks ago, and worked well.

@amosalb Thank you for reporting this, and apologies. Could you please give an example of the test filename and path, which imports it has if any, and which Python test framework you expected it to use. In the interim, please try using `g:test#python#runner`

amosalb:
I believe the issue started with PR #785, but I only realized this after commenting here. The file name with the path is `${workspaceDirectory}/tests/unit/test_app.py`. The test imports `os` and `TestCase` from the `unittest` module, which is the framework I'm using in this project. I will attempt to run the tests using `g:test#python#runner` and provide you with feedback. I haven't set it up yet because `unittest` is the default runner for Python in vim-test.

amosalb:
I made an attempt with the variable runner and worked. Thank you!

codeinabox:
> I made an attempt with the variable runner and worked. Thank you!

That's great! I am still keen to fix the issue you encountered, 

> I believe the issue started with PR #785, but I only realized this after commenting here. The file name with the path is `${workspaceDirectory}/tests/unit/test_app.py`. The test imports `os` and `TestCase` from the `unittest` module, which is the framework I'm using in this project. I will attempt to run the tests using `g:test#python#runner` and provide you with feedback. I haven't set it up yet because `unittest` is the default runner for Python in vim-test.

To confirm, does the import look like this?

```python
from unittest import TestCase
```

amosalb:
Yes, it does

codeinabox:
Thank you for clarifying, I've got a fix incoming https://github.com/vim-test/vim-test/pull/786

codeinabox:
@amosalb does this now work for you if you don't set `g:test#python#runner`?

amosalb:
> @amosalb does this now work for you if you don't set `g:test#python#runner`? 

Yes, it does. I'm running without that variable setted up and it's working fine. Thank you for your willingness and prompt response.

PR #785, opened 2024-01-19
feat: pytest import detection

codeinabox:

### DIFF
--- a/autoload/test/python.vim
+++ b/autoload/test/python.vim
@@ -3,6 +3,10 @@ let test#python#patterns = {
   \ 'namespace': ['\v^\s*class (\w+)'],
 \}
 
+function! test#python#has_import(file, module) abort
+  return match(readfile(expand(a:file)), "import " . a:module) != -1
+endfunction
+
 function! test#python#executable() abort
   if executable('python3')
     return 'python3'
--- a/autoload/test/python/pytest.vim
+++ b/autoload/test/python/pytest.vim
@@ -7,7 +7,7 @@ function! test#python#pytest#test_file(file) abort
     if exists('g:test#python#runner')
       return g:test#python#runner ==# 'pytest'
     else
-      return match(readfile(expand(a:file)),"import unittest") == -1 && (executable("pytest") || executable("py.test"))
+      return !test#python#has_import(a:file, 'unittest') && (executable("pytest") || executable("py.test"))
     endif
   endif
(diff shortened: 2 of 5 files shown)


matt-gardner:
After this commit, I ran into consistent errors with `Not a test file`, where before this commit it was working just fine.  The issue is that it's not uncommon to have tests that are run with `pytest` that don't import `pytest` anywhere in the file, so the check implemented in this PR makes things that were previously working fail.  It took me a fair amount of digging to track down why my tests would no longer run to this PR.

Overall I think it's not unreasonable to have the default behavior be to not run a test if there's no runner set and you can't unambiguously determine what runner to use.  But the error message was "Not a test file", which is wrong, and the language-specific documentation for python says "By default the first available will be chosen", which is no longer true.

If it's more helpful to open this as an issue instead of putting it as a comment here, I can do that too.

codeinabox:
> After this commit, I ran into consistent errors with `Not a test file`, where before this commit it was working just fine.  The issue is that it's not uncommon to have tests that are run with `pytest` that don't import `pytest` anywhere in the file, so the check implemented in this PR makes things that were previously working fail.  It took me a fair amount of digging to track down why my tests would no longer run to this PR.
> 
> Overall I think it's not unreasonable to have the default behavior be to not run a test if there's no runner set and you can't unambiguously determine what runner to use.  But the error message was "Not a test file", which is wrong, and the language-specific documentation for python says "By default the first available will be chosen", which is no longer true.
> 
> If it's more helpful to open this as an issue instead of putting it as a comment here, I can do that too.

Thank you for reporting this. Does your project have a pytest.ini file? It's on my to do list to use it to determine the runner.

matt-gardner:
We don't have a `pytest.ini` file, but we do have a `[tool.pytest.ini_options]` section in our `pyproject.toml`.

codeinabox:
> We don't have a `pytest.ini` file, but we do have a `[tool.pytest.ini_options]` section in our `pyproject.toml`.

@matt-gardner not forgotten about this, just been busy. In the interim, you should be able to use `g:test#python#runner`

matt-gardner:
@codeinabox no worries, nothing's blocking me, I did the `g:test#python#runner` fix before ever commenting here.  I was just giving a heads up about the confusion, as it took me a while to figure out that that was the right fix and trace it here.

PR #787, opened 2024-01-24
Update minitest.vim to use ruby args by default for when runner name does not contain ruby/rake

dewyze:
If you have a strategy that uses `bin/test` to run, it will not match against 'ruby' or 'rake'.

My suggestion is we fall back to ruby when neither is present. While this does presume ruby as the default, the alternative is it not working.

I'd also be happy to make this a configuration option instead if that makes more sense.

**NOTE:** Regarding these checkboxes:
- There are no fixtures or specs for using minitest with ruby in general, just for rake.
- Don't think the README needs to be updated
- I don't see a need for an update in doc/text.txt.


codeinabox (approved):
Thank you for this, a fall back is better than than a configuration option, as the ethos of vim-test is zero configuration required

PR #794, opened 2024-04-20
Allow setting a dynamic project_root

gmartsenkov:
I wanted to dynamically set the project_root and found the easiest way to do that is by allowing it to be a function. That way I can very flexibly decide where the project root is.

### DIFF
--- a/README.md
+++ b/README.md
@@ -448,6 +448,15 @@ different working directory for running tests:
 let test#project_root = "/path/to/your/project"
 ```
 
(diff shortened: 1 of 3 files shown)


codeinabox (approved):
Thank you for this!

Potentially, this means that we could add mono-repo support by leveraging the [FindRootDirectory() in vim-rooter](https://github.com/airblade/vim-rooter?tab=readme-ov-file#using-root-finding-functionality-in-other-scripts)

PR #796, opened 2024-04-26
Add split pane percent to WezTerm strategy

emilford:
The `split-pane` command creates a split that's 50% of the screen width when split using `--right`. This may be wider than some preferred. This adds a `split_percent` option to customize the width of the split pane.


codeinabox (approved):
Thank you!

PR #800, opened 2024-05-19
Support gotest args

roychoo:
specifically
-args
    Pass the remainder of the command line (everything after -args)
    to the test binary, uninterpreted and unchanged.
    Because this flag consumes the remainder of the command line,
    the package list (if present) must appear before this flag.

https://pkg.go.dev/cmd/go/internal/test#pkg-index

btw, i am not very sure where to write the docs as this is very go specific

### DIFF
--- a/README.md
+++ b/README.md
@@ -398,6 +398,11 @@ let test#rust#cargotest#test_options = {
 \}
 ```
 
+The gotest runner let you specify the -args argument as follows:
+```vim
+let test#go#gotest#args = 'a=b'
+```
+
 ### Vim8 / Neovim terminal position
 
 Both the `neovim` and `Vim8 Terminal` strategy will open a split window on the
--- a/spec/gotest_spec.vim
+++ b/spec/gotest_spec.vim
@@ -149,4 +149,20 @@ describe "GoTest"
 
     Expect g:test#last_command == 'go test ./...'
   end
+
+  it "runs test with args"
+    view +5 normal_test.go
+    let g:test#go#gotest#args = "a=b"
+    TestNearest
+
+    Expect g:test#last_command == 'go test -run ''TestNumbers$'' ./. -args a=b'
+  end
+
+  it "runs test suite with args"
+    view +14 build_tags_test.go
+    let g:test#go#gotest#args = "a=b"
+    TestSuite
+
+    Expect g:test#last_command == 'go test ./... -args a=b'
+  end
 end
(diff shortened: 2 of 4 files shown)


spec/gotest_spec.vim
```diff
@@ -149,4 +149,20 @@ describe "GoTest"
 
     Expect g:test#last_command == 'go test ./...'
   end
+
+  it "runs test with args"
+    view +5 normal_test.go
+    let g:test#go#gotest#args = "-args a=b"
```
codeinabox, line 155: Given that the option is called `args` could we make it automatically add `-args` so you could just do the following?
 
```suggestion
    let g:test#go#gotest#args = "a=b"
```

roychoo: sure. will do

roychoo: done

roychoo:
@codeinabox hi, is there any changes that you would still like? and regarding the documentation, is there any location that you want me to write this `option` down?

codeinabox:
> @codeinabox hi, is there any changes that you would still like? and regarding the documentation, is there any location that you want me to write this `option` down?

Thanks for asking, I've updated the documentation structure with #799, so you if you add it below the cargo test option documentation that would be great

roychoo:
@codeinabox done, let me know if it is sufficient

roychoo:
@codeinabox done

README.md
```diff
@@ -361,6 +361,7 @@ If you want some CLI options to stick around, you can configure them in your
 ```vim
 let test#ruby#minitest#options = '--verbose'
 let test#javascript#denotest#options = '--quiet'
+let test#go#gotest#args = 'a=b'
```
codeinabox, line 364: Instead of adding it here, could you add some documentation like the following around line 400


> The gotest runner let you specify the `-args` argument as follows:
> 
> ```vim
> let test#go#gotest#args = 'a=b'
> ```

codeinabox (changes requested):
Almost there, I just want to keep custom options separate from the standard ones in the documentation to avoid confusion.

Could you also update `doc/test.txt` and add some documentation about the option after line 712

README.md
roychoo: ok done. let me know if is ok.

README.md
```diff
@@ -398,6 +398,12 @@ let test#rust#cargotest#test_options = {
 \}
 ```
 
+The gotest runner let you specify the -args argument as follows:
+```vim
+let test#go#gotest#args = 'a=b'
+
```
codeinabox, line 404:
```suggestion
let test#go#gotest#args = 'a=b'
```

codeinabox, line 404: Could you remove this blank line please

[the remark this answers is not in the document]
roychoo: done

codeinabox (approved):
Thank you for all your work on this!

roychoo:
@codeinabox thanks for the review too!

PR #801, opened 2024-05-28
Add dispatch.vim Spawn strategy

### LINKED ISSUE #756: Feature: support for dispatch Spawn
I have been using Vim8 with vim-test for a while using the default basic strategy which is great. I wanted to run my tests non-blocking while still be able to navigate the code, so I tried Dispatch strategy. That allows me to continue to navigate the code, but now I can no longer stop at breakpoints to debug. The Dispatch recommendation is to use Spawn instead of Dispatch as the command.
https://github.com/tpope/vim-dispatch/issues/319

Could vim-test add support for this?
I read through the other options and it wasn't clear that any of the other strategies that are supported would meet desire for non-blocking, debuggability / interactive term. I tried asyncrun but the debugger wouldn't work there either.

bnadlerjr:
dispatch.vim supports the `Spawn` and `Spawn!` commands in addition to the `Dispatch` variations. This adds a strategy to also support `Spawn`.

Closes #756.


codeinabox (approved):
Thank you!

PR #808, opened 2024-07-29
Cargo feature/allow named packages

paulodiovani:
Updates `cargotest` and `cargonexttest` scripts to read `pakage.name` from `Cargo.toml` instead of relying only on the directory name.

Fixes https://github.com/vim-test/vim-test/issues/807

  ^ doesn't look like it is necessary

### DIFF
--- a/autoload/test/rust/cargonextest.vim
+++ b/autoload/test/rust/cargonextest.vim
@@ -114,6 +114,17 @@ function! s:test_namespace(filename) abort
           echo
           let l:package = l:modules[idx]
           let l:modules = l:modules[idx+1:]
+          " use package name, if present
+          let l:section = ''
+          for line in readfile(l:cargo_toml)
+            if line =~ '\[.\+\]' | let l:section = line | endif
+              if l:section == '[package]' && line =~ 'name'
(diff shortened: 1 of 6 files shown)


autoload/test/rust/cargonextest.vim
```diff
@@ -114,6 +114,17 @@ function! s:test_namespace(filename) abort
...
+          " use package name, if present
+          let l:section = ''
+          for line in readfile(l:cargo_toml)
+            if line =~ '\[.\+\]' | let l:section = line | endif
+              if l:section == '[package]' && line =~ 'name'
+              let l:package = matchstr(line, '"\zs[^"]*\ze"')
+
+              break
+            endif
+          endfor
```
paulodiovani, line 126: This was my attempt to read the package name from a `Cargo.toml` using only RegExp.

If anyone knows of a better way, please leave a suggestion. ;)

codeinabox: I think it's fine to use RegExp to parse it as you have added tests around this

codeinabox (approved):
This is great, thank you very much!

PR #810, opened 2024-08-12
Run `pest` via `sail` when configured

timacdonald:
This PR adds support for running Pest test suites via Laravel Sail automatically.

Sail is already detected for PHPUnit testsuites. This uses the same mechanism.

I have not updated the readme or docs as this feels more like an enhancement than a feature worth documenting.

### DIFF
--- a/spec/sail_spec.vim
+++ b/spec/sail_spec.vim
@@ -11,7 +11,7 @@ describe "Laravel Sail"
 
   after
     call Teardown()
-    !rm -rf vendor
+    !rm -rf vendor docker-compose.yml
     cd -
   end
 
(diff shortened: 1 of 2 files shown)


spec/sail_spec.vim
```diff
@@ -11,7 +11,7 @@ describe "Laravel Sail"
 
   after
     call Teardown()
-    !rm -rf vendor
+    !rm -rf vendor docker-compose.yml
```
timacdonald, line 14: Running the testsuite locally, this file was not cleaned up. Just making sure we have a fresh slate for every test.

timacdonald:
Thank you!

PR #814, opened 2024-09-08
Run test cases prefixed with t_ with CommonTest runner

dmitrivereshchagin:
Also simplify test pattern customization for CommonTest runner.

### DIFF
--- a/autoload/test/erlang.vim
+++ b/autoload/test/erlang.vim
@@ -1,4 +1,0 @@
-let test#erlang#patterns = {
-  \ 'test':      ['\v^\s*(test_\w*)\('],
-  \ 'namespace': [],
-\}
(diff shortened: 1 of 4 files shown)


autoload/test/erlang.vim
```diff
@@ -1,4 +0,0 @@
-let test#erlang#patterns = {
-  \ 'test':      ['\v^\s*(test_\w*)\('],
-  \ 'namespace': [],
-\}
```
codeinabox, line 4: This shouldn't need to be deleted

dmitrivereshchagin: I think it's OK to delete it.

This variable name is misleading.  The variable isn't used (and cannot be) in all Erlang runners.  It also requires to specify namespace key that cannot be used and is simply ignored.  And it's not documented.

I also noticed that similar variables are often defined unconditionally along with functions in the same autoload script.  That is, in such a way that they are difficult to override.

Is there any reason other than minor BC break to keep this variable?

codeinabox: Alright, you've convinced me, and I've noticed that not every file type has or needs this

codeinabox (approved):
Thank you!

PR #817, opened 2024-10-01
Allow configuring behat executable

TRPB:

There's no tests for the other executable options so didn't know how you wanted to handle that

### DIFF
--- a/README.md
+++ b/README.md
@@ -678,6 +678,12 @@ Similarly if you'd prefer to use an alternate runner such as the [Laravel artisa
 let test#php#phpunit#executable = 'php artisan test'
 ```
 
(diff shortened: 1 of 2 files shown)


codeinabox:
Thank you for raising this pull request, though [vim-test already supports custom executables](https://github.com/vim-test/vim-test?tab=readme-ov-file#executable).

I just tested this myself and was able to override the executable. Are you unable to do this without these changes?

TRPB:
it doesn't work for me, it tries to run `vendor/bin/behat` which exists but doesn't work because I don't have the same php config locally as in the docker environment I'd like to use. 

The existing phpunit executable override (which works) is handled the same way as this PR: https://github.com/vim-test/vim-test/blob/master/autoload/test/php/phpunit.vim#L52

Issue #819, opened 2024-10-21
Add support for xmake runner


dkaszews:
I will probably end up implementing it myself, so using this issue for notes and to judge interest.

[Xmake](https://xmake.io/#/) is a cmake-like build utility that I enjoy using for its lua-based scripting, terseness and support for bunch of package managers. As example, here is all that is needed to set up a TDD C++ project with gtest:

```lua
-- xmake.lua
add_requires("gtest >= 1.14.0")

set_languages('c++17')
set_warnings('all', 'extra', 'pedantic', 'error')

target('my_app')
    set_kind('binary')
    add_files('src/*.cpp')
    add_includedirs('include')

target('my_test')
    set_kind('binary')
    add_files('test/*.cpp')
    add_includedirs('include')
    add_packages('gtest')
    add_tests('default')
```

Then run it with `xmake test -v` (`-v` makes it print gtest output if tests fail).

dkaszews:
I was playing around with this and hit a bit of a snag, as the current architecture picks the first runner whose `test_file` returns true. Since I want to support any file in a directory containing `xmake.lua` (or subdirectory, recursively), I am being non-deterministically (unordered hashset) overridden by other runners.

It also seems like I am going against established rules by creating a runner that doesn't really care about language of particular file, just where it is located. For now, the best idea I have is to create a "many" or "any" pseudo-language, which has higher priority than anything else.

dkaszews:
I think this issue will be mostly covered by #820 with simple `{ "command": "xmake test" }` placed in `.vimtest.json` at project root. If that gets merged, then I will put this one on hold unless there is additional interest for proper support of xmake.

dkaszews:
I guess, I've just been committing `.vimtest.json` to my xmake repos, so we can close this until xmake gains enough popularity that a dedicated runner makes sense.

PR #821, opened 2024-10-27
Implement generic command runner using .vimtest.json

### LINKED ISSUE #820: Add support for generic command runner
I will probably end up implementing it myself, so using this issue for notes and to judge interest.

I would like to be able to quickly configure a custom command to test a project without implementing a whole strategy. It could be implemented by looking for a `.vimtest.sh` file (or other extensions, such as `.bat`, `.ps1`, etc.), and executing it if found.

dkaszews:
Resolves #820


codeinabox:
Thank you for looking into this, could you please add an example to the fixtures folder and an accompanying spec. I am most interested in how this will behave. Will it only support TestFile and TestSuite?

dkaszews:
@codeinabox This is a draft for sharing code, there is ongoing discussion in the issue how to best approach it, what usecase should be covered, etc. Once we settle on a particular design, I will make sure to provide documentation, examples and test cases.

paolooo:
@dkaszews Sounds great. I'm not entirely sure how to set this up, but after checking the CI, it looks like I need ruby v3.2 and `vim-flavor` plugin manager to run the specs, right?

dkaszews:
@paolooo You don't need to run specs because I have not yet written any, but you can checkout the branch to test it manually. Cd into wherever you downloaded vim-test (depends on your system and vim/neovim, Google should help), then `git switch https://github.com/dkaszews/vim-test.git generic_command` and restart vim.

codeinabox:
A couple of questions:

- Would `TestSuite` or `TestFile` work from any file? Will it be higher or lower priority than the existing runners?
- Am I correct in assuming that `TestNearest` won't be supported due to the complexity?

dkaszews:
@codeinabox

> Would TestSuite or TestFile work from any file?

For now all `:Test*` execute the same command, to keep it simple as first implementation, see what people actually need. So all of `TestFile` and `TestSuite` and `TestNearest` will invoke the exact same command defined in JSON. It will work for all files that are in the same directory or subdirectory of `.vimtest.json`, basic usecase is to put it in project root. 

> Will it be higher or lower priority than the existing runners? 

This runner is intended as an override, so it has higher priority than anything else. I achieved it by assigning it to language `_` (`*` would look nicer but is not a good folder name) and sorting runners alphabetically. If no `.vimtest.json` is found in current file's directory or its parents, then it looks at other runners as usual.  

> Am I correct in assuming that TestNearest won't be supported due to the complexity?

I picked JSON instead of raw script so it can be easily extended in the future with backwards compatibility. I have some ideas how to support `TestFile` and `TestNearest` but they require a lot more work and testing, so they won't be implemented here.

codeinabox:
> This runner is intended as an override, so it has higher priority than anything else. I achieved it by assigning it to language `_` (`*` would look nicer but is not a good folder name) and sorting runners alphabetically. If no `.vimtest.json` is found in current file's directory or its parents, then it looks at other runners as usual.

So in summary, if you use `.vimtest.json` you will be responsible for configuring everything because it overrides the default zero config approach?

dkaszews:
Correct, it is mostly so that if you can run custom commands without implementing a whole new runner, while still integrating with existing strategies and mappings. Also, by storing as file inside a repo, other commiters do not have to update their vimrc, similar to how `.gitignore`, `.clangformat`, `.vimspector.json` and many others.

As example usecase, you can see #819. I plan to write a runner for it anyways, but it will use at least some of code here as a base, since it also looks for a `xmake.lua` file instead of just at currently open file. Similar approach could be to look for `CMakeLists.txt` to run commands in `cpp#catch2` runner even if you are currently looking at a non-C/C++ file (cmake, assembly, template, custom config).

codeinabox:
Let me mull this over. My reservation at this stage is this a big departure from the vim-test philosophy of zero config, and I'm worried it might detract from improving runner detection. The custom commands you are creating, do they still use standard test runners underneath? If so could we add support for those?

> Correct, it is mostly so that if you can run custom commands without implementing a whole new runner, while still integrating with existing strategies and mappings.

If it's simply a case of a custom command, instead of a generic runner, could we introduce a new command say `TestCommand`?

paolooo:
I think we should prioritize the current `vim-test` runners. If vim-test doesn't support, then, will use the command. For example:

```json
{
}
```

dkaszews:
@paolooo Issue is that the current runners may think they support it, but don't really. For example, just because you have a python file open, doesn't mean you actually want to run python tests, as it might be just part of a build script for a C# project. If somebody explicitly tells vim-test which command they want to run, then they should be treated as an authority.

paolooo:
That makes sense, that’s a scenario to consider.

In my case, it was running `vue-cli-service` instead of `vitest`, so overriding the JavaScript executable works for me. I think the following configuration will work for me:

```json
{
  "javascript": {
    "executable": "vitest"   // instead of command
  }
}
```

I think `executable` make sense for me because of function `test#javascript#vitest#executable()`.

dkaszews:
I don't think executable is part of the base API. In general, if you want to configure something within an existing runner for a well-known framework, you can either do it via some global setting, or add a pull request to enable such configuration.

dkaszews:
@codeinabox Have you had time to consider this? While the code could use bit more care, I have been using this branch for the past two weeks at work, with projects like:

* MSVC projects, running UT with `mstest.exe`
* FW assembly projects, running multi-step compiler scripts to check if code still compiles
* Bash scripts, comparing output with golden standard
* Performance regression testing using [hyperfine](https://github.com/sharkdp/hyperfine)

I simply like the ability to run whatever counts as a "test" at given moment with a binding I am already used to.

codeinabox:
Hello @dkaszews thank you for your hard work and patience on this, as I do agree a custom test command would be useful in these instances. However, my concern of this overriding the zero config mantra of the plug-in.

Instead of having a `.vimtest.json`, is an alternative approach to add mappings to a project specific config ie `.vimrc`? Either way, you need to maintain custom configuration per project.

If we do go with the `.vimtest.json` approach, I'd want to make it clear in the documentation that users are opting out of the zero configuration approach.

dkaszews:
@codeinabox Thank you for consideration. My main issue with vimrc is the syntax and the increased difficulty of sharing configs appropriate for given project.

To illustrate the folder based approach, at work I have setup which looks something like this:

```
project_root/
  |- .vimtest.json  # ./compile_solution.sh
  |- compile_solution.sh
  |- ... 
  |- fw/
    |- .vimtest.json  # ./compile_firmware.sh
    |- compile_firmware.sh
    |- ... 
  |- tests/
    |- .vimtest.json  # ./run_tests.py
    |- run_tests.py
    |- ... 
```

So if I run `:TestFile` while editing a FW file, it will compile FW, for a test file it will run tests, for anything else it will run the entire solution (compile and test everything, plus flash binary on target).

If I wanted to achieve something similar with a `.vimrc`, I would have to do check for different prefixes of the file (in proper order, so the longer ones have priority) and make sure to update it if the layout changes or I clone the repository in another location (e.g. to have two different copies to compare while investigating a bug). Plus, my colleagues would have to implement the same thing for their workspaces.

codeinabox:
> This PR just aims to build what I feel to be common pattern (as seen in other plugins and tools) into the tool, so that people don't have to reimplement it themselves.

Could you please provide more context - are you suggesting that people will start committing `.vimtest.json` into their repositories?

Do you envisage this approach is for expert users with custom testing requirements, and the existing zero config approach is recommended for everyone else?

dkaszews:
In short, yes, I expect expert users to keep a `.vimtest.json` in their repositories, if their repository does something which cannot fit under any existing runner. Same as they would commit a `.gitignore`, `.clangformat` or other project description files.

codeinabox:
Great, it's important we both agree this is a feature for expert users. If you could add some documentation and specs for this generic runner, then I can get this feature merged in.

dkaszews:
@codeinabox I found some time to get back to this PR, added specs and docs, did bit of cleanup and fixed it for Vim (NeoVim accepts `read_file()->json_decode()`, Vim requires an explicit `join()` in the middle).

Feel free to review and merge at your pace.

dkaszews:
@paolooo Hi, would you like to give this version a spin? API did not change, but it should work on Vim now.

If you wanted to run specs for some reason, you need to install Ruby, then `gem install vim-flavor`. I also needed to manually add `~/.gem/bin` to `$PATH`. Best way to run them is using `vim-test` itself 😀, `:TestFile` for single or `:TestSuite` for all.

paolooo:
Awesome @dkaszews! Will check this out. 🎉

README.md
```diff
@@ -754,6 +754,25 @@ endfunction
 let g:test#custom_alternate_file = function('CustomAlternateFile')
 ```
 
+## Overriding test commands
+
+This is considered an advanced feature, subject to active development and further changes.
```
codeinabox, line 759: It needs to be made clear that this feature overrides the zero configuration approach

```suggestion
This is considered an advanced feature, subject to active development and further changes. It overrides the zero configuration approach, and requires you to manually configure the test runners.
```

doc/test.txt
```diff
@@ -840,6 +841,26 @@ endfunction
 let g:test#custom_alternate_file = function('CustomAlternateFile')
 <
 
+OVERRIDING TEST COMMANDS                        *test-overriding-test-commands*
+
+This is considered an advanced feature, subject to active development and further changes.
```
dkaszews, line 846:
```suggestion
This is considered an advanced feature, subject to active development and further changes.
It overrides the zero configuration approach, and requires you to manually configure the test runners.
```

README.md
dkaszews: Sure, done and same in doc

codeinabox:
> @paolooo Hi, would you like to give this version a spin? API did not change, but it should work on Vim now.

Would you like me to hold off reviewing this or is it ready to be reviewed?

paolooo:
Hi @dkaszews, thanks for the update! I'm having trouble figuring out how to test this. Are there any documentation or instructions on setting up this new version for testing?

dkaszews:
@paolooo If you mean how to use this code before it gets merged, you have to manually check it out:

1. Find out where vim-test is installed, easiest is to ask vim where some function or command was defined:
    ```vim
    :verbose command TestFile
    :verbose function test#exists
    ```
    and look for "last set from"
1. Cd into that directory
1. At the top of PR just under the title you have source branch `dkaszews:generic_command`, click on it to open the remote and copy URL from the green "clone" button
1. Run following commands:
    ```sh
    git remote add dkaszews https://github.com/dkaszews/vim-test.git
    git fetch dkaszews
    git switch -c generic_command dkaszews/generic_command
    ```
1. Restart vim to load the new version
1. If you want to switch back to mainline, run `git switch master` in the same directory

codeinabox (approved):
Thank you for your work on this!

PR #825, opened 2024-10-31
Fix neovim_sticky strategy in Linux PowerShell

### LINKED ISSUE #824: neovim_sticky strategy does not work with PowerShell
It appears that `chansend` does not work properly with PowerShell, as it only executes commands on `\r` and not `\n`, the latter being used implicitly when using a list of commands. Since bash works with either, this should be fixable by explicitly joining the list of commands with `\r`:

```diff
@@ -134 +134 @@ function! test#strategy#neovim_sticky(cmd) abort
-  call chansend(l:buffers[0].variables.terminal_job_id, l:cmd)
+  call chansend(l:buffers[0].variables.terminal_job_id, join(l:cmd, "\r"))

```

Requires more testing on multiple systems and different shells to see if nothing breaks.

dkaszews:
Fixes #824 , bug fix no docs update needed.

### DIFF
--- a/autoload/test/strategy.vim
+++ b/autoload/test/strategy.vim
@@ -131,7 +131,8 @@ function! test#strategy#neovim_sticky(cmd) abort
     let l:win = [s:neovim_reopen_term(l:buffers[0].bufnr)]
   endif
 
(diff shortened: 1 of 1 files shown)


codeinabox (approved):
Thank you!

PR #828, opened 2024-12-17
Fix term_position usage in s:neovim_reopen_term

Danthewaann:
If `test#neovim#term_position` is set to something like `botright 10` to setup a terminal that is 10 in height and we need to reopen the terminal since it isn't visible, the current `s:neovim_reopen_term()` function doesn't correctly re-open the terminal with the appropriate height. This PR addresses this by running something like `botright 10 new | buffer [bufnr]` which sets the height correctly instead of `botright 10 sbuffer [bufnr]` which doesn't.


codeinabox (approved):
Thank you for your contribution and the detailed explanation!

PR #829, opened 2024-12-18
Maven: fix multimodule when one submodule depends on another

jqno:
Running a single test or test suite in a multimodule Maven project doesn't work when the test is in a submodule `B` that depends on another submodule `A`. In these cases, we need to add:

- `-am` so Maven (re)builds module `A` as well as `B`, if necessary
- `-Dsurefire.failIfNoSpecifiedTests=false` so the build doesn't fail when the specified test from module `B` doesn't exist in `A` (only in case of running unit tests with surefire)
- `-Dfailsafe.failIfNoSpecifiedTests=false` so the build doesn't fail when the specified test from module `B` doesn't exist in `A` (only in case of running integration tests with failsafe)

If the test is in `A`, these parameters aren't necessary, but it doesn't matter if they're there. If the test is in `B`, Maven fails because it tries to find the test in `A` and can't find it there. So we'll just add the parameters in all multi-module cases.

I've added a dependent module in both `sample_maven_junit3_multimodule_project` and `sample_maven_junit5_multimodule_project`, even if it's not strictly necessary for `vim-test`'s tests, so the underlying issue can be reproduced.

Here's the error in question:

```
$ mvn test -Dtest=org.vimtest.math.# -pl sample_module2

[INFO] Scanning for projects...
[INFO] 
[INFO] ------------------< org.vimtest.calc:sample_module2 >-------------------
[INFO] Building sample_module2 1.0-SNAPSHOT
[INFO]   from pom.xml
[INFO] --------------------------------[ jar ]---------------------------------
[WARNING] The POM for org.vimtest.calc:sample_module:jar:1.0-SNAPSHOT is missing, no dependency information available
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  0.251 s
[INFO] Finished at: 2024-12-18T15:56:26+01:00
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal on project sample_module2: Could not resolve dependencies for project org.vimtest.calc:sample_module2:jar:1.0-SNAPSHOT
[ERROR] dependency: org.vimtest.calc:sample_module:jar:1.0-SNAPSHOT (compile)
[ERROR] 	Could not find artifact org.vimtest.calc:sample_module:jar:1.0-SNAPSHOT
[ERROR] 
[ERROR] -> [Help 1]
[ERROR] 
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[ERROR] Re-run Maven using the -X switch to enable full debug logging.
[ERROR] 
[ERROR] For more information about the errors and possible solutions, please read the following articles:
[ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/DependencyResolutionException
```

With this command it succeeds:
```
$ mvn test -Dtest=org.vimtest.math.# -Dsurefire.failIfNoSpecifiedTests=false -am -pl sample_module2
```
---

- [X] Update the Vim documentation in `doc/test.txt`

### DIFF
--- a/autoload/test/java/maventest.vim
+++ b/autoload/test/java/maventest.vim
@@ -20,17 +20,23 @@ function! test#java#maventest#build_position(type, position) abort
   " ex:  mvn test -Dtest com.you.pkg.App$NestedClass#test_method
   " ex:  mvn test -Dtest com.you.pkg.App#test_method
   " ex:  mvn test -Dtest com.you.pkg.App\*           (catches nested test-classes)
-  " ex:  mvn test -Dtest com.you.pkg.App\* -pl module_name
-  " ex:  mvn test -Dtest com.you.pkg.App#test_method -pl module_name
+  " ex:  mvn test -Dtest com.you.pkg.App\* -Dsurefire.failIfNoSpecifiedTests=false -am -pl module_name
+  " ex:  mvn test -Dtest com.you.pkg.App#test_method -Dsurefire.failIfNoSpecifiedTests=false -am -pl module_name
 
   if exists('g:test#java#maventest#test_cmd')
     let test_cmd = g:test#java#maventest#test_cmd
   else
-    let test_cmd = 'test -Dtest='
-
     if filename =~# 'IT\|ITCase\|Integration$' && a:type =~# '^nearest\|file$'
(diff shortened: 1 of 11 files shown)


codeinabox (approved):
Thank you! Love the thorough and well tested approach

PR #831, opened 2025-01-04
Fix cargo test handling of `src/bin` files

alexgenco:
This is a special directory for projects that author multiple binaries. When focusing tests in these directories, `bin::` should be omitted from the module path, and the `--bin` argument should be set to the appropriate binary name (derived from the path).

See:
https://doc.rust-lang.org/cargo/reference/cargo-targets.html#binaries


codeinabox (approved):
Thank you!

PR #841, opened 2025-04-30
Add `uv` support for python runners

### LINKED ISSUE #840: Support for UV for `pytest` and `djangotest`?
https://github.com/astral-sh/uv

It's a new package manager for `python` that very quickly gains traction. I think the easiest way to detect it would be with `uv.lock`.
And then to run `uv run` instead of the usual `python`.

juliancoffee:
Fixes #840 

I mostly mimicked https://github.com/vim-test/vim-test/pull/538, but let me know if I need to tweak something.

P.S. I went by grepping `pipenv` and added `uv` in the same places, and I see `pipenv` is supported for more runners than just `pytest` and `djangotest`, or so it seems at least 😅 
P.P.S off-topic, but just wanted to say that it was a pleasure finding this plugin, not just because it's the first one that works, but also because you're really supportive of contributors, so thank you!

### DIFF
--- a/README.md
+++ b/README.md
@@ -499,8 +499,9 @@ and will use `poetry run pytest` if it detects a `poetry.lock`. The pyunit and n
 runner supports [pipenv](https://github.com/pypa/pipenv) as well and will
 respectively use `pipenv run python -m unittest` or `pipenv run python -m nosetests`
 if there is a `Pipfile`. It also supports [pdm](https://pdm.fming.dev/) as well and
-will use `poetry run pytest` if there is a `pdm.lock` file.
-
+will use `poetry run pytest` if there is a `pdm.lock` file. As well as [uv](https://github.com/astral-sh/uv) and will use `uv run pytest` if there is `uv.lock` file.
+All runners except `djangotest` support uv and will use `uv run` if there's a
+`uv.lock` file.
 #### Java
 
 For the same reason as Python, runner detection works the same for Java. To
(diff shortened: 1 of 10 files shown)


juliancoffee:
Ok, I just tried this, and it seems that while it works, it does break the runner detection in a way, so if I run it without first sourcing the `activate` script, it uses `uv` but with `unittest`, so `uv run python unittest`, but if I run it from virtualenv, it correctly runs with `uv run python manage.py test`

I wonder how others, Poetry/Pipenv/etc, work with this?

codeinabox:
Thank you for raising this PR and the kind words. How can I help and test this?

autoload/test/python/djangotest.vim
```diff
@@ -39,6 +39,8 @@ function! test#python#djangotest#executable() abort
     let pipenv_prefix = "pipenv run "
   elseif filereadable("poetry.lock")
     let pipenv_prefix = "poetry run "
+  elseif filereadable("uv.lock")
+    let pipenv_prefix = "uv run "
```
codeinabox, line 43: If this doesn't work with django I'd remove this

```suggestion
```

codeinabox (changes requested):
For now, could you please remove the change for django, remove `djangotest_uv_spec.vim`, and create specs for mamba, nose, pytest and pyunit. I'd keep them simple and just test `TestFile`.

autoload/test/python/djangotest.vim
juliancoffee: I guess I'll do that

juliancoffee:
@codeinabox done 🙌 
It's a shame that I couldn't get Django to work, but I'm glad that others work for you :)

P. S. I hope I did spec files right, but let's see what CI thinks about them 😅

codeinabox (changes requested):
Almost there, you need to update the tests to remove the uv.lock afterwards.

juliancoffee:
@codeinabox CI should be good now; I've also updated the docs as you requested :)

PR #842, opened 2025-05-30
fix: correct catch2 typo

liamvdvyver:


codeinabox (approved):
Thank you and well spotted!

PR #847, opened 2025-06-17
Remove useless echo of nearest test info for PHPUnit

dietrichm:
The information for the nearest PHPUnit test is printed and triggers a hit-enter prompt in some cases. I don't see the value of this being printed apart for debugging purposes.


codeinabox (approved):
Well spotted and thank you!

PR #851, opened 2025-08-19
Add Nearest support to vim-themis (vimspec style)

bootleq:
Themis can use `--target pattern` to filter tests, which can be used as the Nearest position.

This PR only handles "vimspec" style, because in "basic" style (where a test is a vim function) it is unreliable to identify tests by merely pattern searching.

- [x] Update the Vim documentation in doc/test.txt

(No need to update README and documentation)


codeinabox (approved):
Thank you for this addition, very much appreciated!

PR #863, opened 2025-09-09
Disable generic runner (.vimtest.json) by default

### LINKED ISSUE #852: Suggest to disable generic runner (.vimtest.json) by default
Hi, I'd like to suggest disabling the generic command runner (`.vimtest.json`) feature by default.
Not only because it's an advanced use case, but more importantly it is too powerful.
A crafted `.vimtest.json` in an upper directory could affect unsuspecting users, causing them to run arbitrary commands unexpectedly.

Currently, explicitly set `g:test#enabled_runners = [some_other_runners]` can disable it.

The planned change can be small: remove `_Generic` from `default_runners`.
Users who need it can still easily opt in by defining it in `custom_runners`.

To provide finer control, it would be good to extend its `test_file()` to respect a whitelist of allowed `.vimtest.json` file paths.

bootleq:
Fix #852 

### DIFF
--- a/README.md
+++ b/README.md
@@ -760,6 +760,12 @@ let g:test#custom_alternate_file = function('CustomAlternateFile')
 
 This is considered an advanced feature, subject to active development and further changes. It overrides the zero configuration approach, and requires you to manually configure the test runners.
 
+Enable the feature by adding this custom runner:
+
+```vim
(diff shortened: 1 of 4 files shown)


dkaszews (approved):
LGTM

codeinabox (approved):

PR #864, opened 2025-09-10
Improve Nearest test for vim-themis (vimspec style)

bootleq:
Hi, this is an enhancement after #851, details:

- Fix vim pattern special characters in test name, they should be properly escaped.
- Prepend "suite" name to test name, this helps to target a test case more accurate.

The produced test names become longer, but I believe it is acceptable, since we don't need to type it by hand.

(No need to update README and documentation)


codeinabox (approved):
Thank you!

PR #871, opened 2025-11-15
Fix triggering of autocmds when exiting terminal

mmrwoods:
When using vimterminal or neovim strategy a `BufDelete` autocmd is used to jump back to the previous window, but this is not compatible with other plugins or configuration that relies on `BufEnter` or `WinEnter` autocmds as they do not get triggered. Fix by triggering nested autocmds.

For example, this broke lightline such that it would appear that the current window was actually the previous window as the lightline update autocmds would not be triggered. There are also examples in the vim docs that would be broken without using `++nested`, e.g. `:help 'cursorcolumn'`

none of the above are applicable here

### DIFF
--- a/autoload/test/strategy.vim
+++ b/autoload/test/strategy.vim
@@ -102,7 +102,7 @@ endfunction
 
 function! test#strategy#neovim(cmd) abort
   call s:neovim_new_term(a:cmd)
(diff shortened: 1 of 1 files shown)


codeinabox (approved):
Thank you!

mmrwoods:
Thanks for merging, again :-)

PR #893, opened 2026-04-07
feat: Checking for multiline import statements in js/ts files

stevematney:

- [x] Update the Vim documentation in `doc/test.txt` (Not needed.)

### DIFF
--- a/autoload/test/javascript.vim
+++ b/autoload/test/javascript.vim
@@ -26,9 +26,12 @@ function! test#javascript#has_package(package) abort
 endfunction
 
 function! test#javascript#has_import(file, import) abort
-  let l:file = expand(a:file)
+  let l:file = substitute(a:file, '\v\\([()$ ])', '\1', 'g')
   if !filereadable(l:file)
-    let l:file = substitute(l:file, '\v\\([()$ ])', '\1', 'g')
+    let l:file = expand(l:file)
+  endif
+  if !filereadable(l:file)
+    return 0
   endif
 
   let l:source = substitute(join(readfile(l:file), "\n"), '\v\_s+', ' ', 'g')
--- a/autoload/test/javascript.vim
+++ b/autoload/test/javascript.vim
@@ -12,8 +12,8 @@ function! test#javascript#has_package(package) abort
   let l:packages = readfile('package.json')
 
   if exists('*json_decode')
-	let l:dict = json_decode(join(packages, ''))
-	return has_key(get(dict, 'dependencies', {}), a:package) || has_key(get(dict, 'devDependencies', {}),  a:package)
+    let l:dict = json_decode(join(packages, ''))
+    return has_key(get(dict, 'dependencies', {}), a:package) || has_key(get(dict, 'devDependencies', {}),  a:package)
   endif
 
   for line in packages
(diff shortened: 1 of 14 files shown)


autoload/test/javascript.vim
```diff
@@ -4,35 +4,200 @@ let test#javascript#patterns = {
...
+    let g:test#javascript#package_cache = {}
+  endif
+endfunction
+
+function! s:get_current_file_search_dir() abort
+  let l:current_file = expand('%:p')
+  return empty(l:current_file) ? getcwd() : fnamemodify(l:current_file, ':h')
+endfunction
+
+function! s:find_file_upward(pattern, Callback = {path -> 1}) abort
```
codeinabox, line 142: What's the purpose and context for this? Is this for handling monorepos? Currently the plug-in assumes that the working directory is root directory of a project containing the usual files ie package.json

stevematney: Yes, this is for monorepos. In those contexts, it's not reliable to assume the base package.json has the correct dependencies, as that's much more likely to appear in the workspace package.json.

Given the intent of the plugin to be as configuration-free as possible, this should more reliably automatically select the correct runner based on the file being tested. It also shouldn't break on repos without workspaces. In that case, it should work exactly the same as it already does.

codeinabox: Is it possible for this pull request the focus is on dealing with multi-line imports and we can split up the work for monorepos and workspaces into different pull requests? I've got some ideas about that and we could discuss them in an issue or a discussion if that works.

stevematney: Sorry for my slowness to reply! I will get to this! I just have been quite busy with other things in the meantime.

stevematney: Done! I have separated this into just the multiline import checks. I've also opened #901 which contains the monorepo improvements.

spec/jest_spec.vim
```diff
@@ -1,14 +1,17 @@
 source spec/support/helpers.vim
 
+let s:repo_dir = getcwd()
+let s:fixture_dir = s:repo_dir . '/spec/fixtures/jest'
+
 describe "Jest"
 
   before
-    cd spec/fixtures/jest
+    execute 'cd ' . fnameescape(s:fixture_dir)
```
codeinabox: What's the reason for this change? What happens if the before and after blocks aren't changed?

stevematney: The reason for this is based on [lines 3-4](https://github.com/vim-test/vim-test/pull/893/changes/BASE..ec64dd13971cd69eeca246f4b51c304731ea5588#diff-76b78aeca7b066ce62ba1a4308de288895642c974d8de13b09131986920c83cdR3-R4), where we get `repo_dir` using `getcwd()`. I don't think we can guarantee that the path doesn't have any special characters that might blow up in an `execute` command (because a repo path could be anywhere on someone's machine), so we have to escape it.

Ultimately, the intent is to `cd` to a full path in order to guarantee we land in the right place, no matter what may have happened in the meantime.

stevematney: If it would help, I can also reset this file to only include relevant changes. The moving directories work was from me getting tests working on my Windows machine. I don't know that these changes are even necessary for that. They're more just a hardening measure.

codeinabox: Thank you, that would be great to keep the changes to a minimum. Thank you for your patience on this.

stevematney: No problem! That should be updated now. I removed some changes from a few files, and it looks like everything remaining is specific to the multiline import detection.

codeinabox (approved):
Thank you for all your work on this!

PR #903, opened 2026-05-06
fix: make TestNearest search both directions for zig

mark2185:

I'm not sure if this is on anyone's wishlist, but I found it odd that `TestNearest` behaved like `TestFile` when there were no tests above the current line. This PR makes it run the nearest test _below_ in case it finds no tests upwards, which is what I'd expect from the _Nearest_ in the name, and if somebody actually wanted to run _all_ of the tests then there's `TestFile`.

This is now probably an inconsistency compared to other test runners, but I haven't used any of them so...

In any case, I'd rather have this closed (but not deleted) in case somebody has the same itch than not having it published anywhere at all.

### DIFF
--- a/autoload/test/zig/zigtest.vim
+++ b/autoload/test/zig/zigtest.vim
@@ -35,5 +35,8 @@ endfunction
 
 function! s:nearest_test(position) abort
   let name = test#base#nearest_test(a:position, g:test#zig#patterns)
+  if name.test_line == -1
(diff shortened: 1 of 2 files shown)


codeinabox:
Thank you for this! Could you please update the spec for this? It should just be a case of adding another example for `TestNearest`

mark2185:
Sure thing! 

~As soon as I fix a small bug that I've stumbled upon (wrong line number when searching downwards because of the line indices).~

EDIT: on second thought, that's a separate issue that does not clash with this change, it'll be addressed in a follow-up PR.

codeinabox (approved):

PR #905, opened 2026-05-06
fix: last test line index

mark2185:

Without  the rewrite of the `for` loop there was an _off-by-two_ error when searching downwards.

Running just the tests without the fix in place:
```
-------- Testing a Vim plugin
spec/zig_spec.vim .. 1/?
not ok 5 - Zig update g:test#last_position with exact line of the test
# Expected g:test#last_position['line'] == 4 at line 4
#       Actual value: 6
#     Expected value: 4
spec/zig_spec.vim .. Failed 1/5 subtests

Test Summary Report
-------------------
spec/zig_spec.vim (Wstat: 0 Tests: 5 Failed: 1)
  Failed test:  5
Files=1, Tests=5,  1 wallclock secs ( 0.01 usr  0.00 sys +  0.05 cusr  0.04 csys =  0.10 CPU)
Result: FAIL
```

The `last_position['line']` is used in `test#visit` to jump to the place where the test was invoked _from_, but I think it makes more sense that `TestVisit` actually visits _the test it ran_.

In case the user ran `TestFile`, it won't update the `last_position['line']`.

~Note: the PR contains the PR #903 because it tests its new functionality (searching downwards) so that one should get merged first.~ Rebased!

Note: `TestVisit` for test runners other than `zig` will still exhibit the current behavior.

### DIFF
--- a/autoload/test/base.vim
+++ b/autoload/test/base.vim
@@ -114,7 +114,6 @@ function! test#base#nearest_test_in_lines(filename, from_line, to_line, patterns
   let test         = []
   let namespace    = []
   let last_indent  = -1
-  let current_line = a:from_line + 1
   let test_line    = -1
   let last_namespace_line = -1
   let is_namespace_with_same_indent_allowed = get(configuration, 'namespaces_with_same_indent', 0)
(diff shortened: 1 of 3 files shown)


mark2185:
@codeinabox I also have a patch that will enable the test runner to run tests that don't have a string literal as a name, but an identifier (`test addOne { ...` that's already in the test file), should I open up a new PR for that or just chuck it in here as a separate commit? 

It's just an additional regex pattern and a spec update.

codeinabox (approved):
Thank you!

PR #909, opened 2026-05-28
fix: escape spaces for busted --filter

vitaly:
Ref: https://github.com/vim-test/vim-test/issues/908

Fixed busted to escape spaces for the `--filter` argument.

Updated `spec/busted_spec.vim` too.

> Note: `(` and `)` magic characters are still broken, other magic characters work fine.


codeinabox (approved):
Thank you for this fix, very much appreciated!

PR #911, opened 2026-07-27
fix: shell-escape filenames with special characters in build_position

### LINKED ISSUE #789: Error if file name contains shell special characters
AppRouter in next.js, for example, allows a path parameter to be specified in the file path.
For example: `/app/user/[id]`
Such a path is recognized as a special pattern by shells such as zsh.
As a result, when you run `:TestFile`, the shell will search for the file, resulting in the following error

Is it possible to have the file name escaped or quoted before passing it to the shell?

mmustafasenoglu:
## Summary

Fixes #789

When test filenames contain shell special characters like `[`, `]`, `$`, `*`, `?`, etc., the shell interprets them as glob patterns or variable references, causing test execution to fail.

**Example (zsh):**
```
$ pytest test_weir[d].py
zsh: no matches found: test_weir[d].py
```

**After fix:**
```
$ pytest 'test_weir[d].py'
# Works correctly - shell treats [d] literally
```

## Approach

Modified `test#base#build_position` in `autoload/test/base.vim` to detect when a filename contains shell special characters and shell-escape it in the returned args.

The fix uses `substitute()` with `\V` (very nomagic) to replace the raw file path with its `shellescape()` equivalent only within args that contain it. This means:

- **Normal filenames** (e.g. `test_class.py`): No change, existing behavior preserved
- **Filenames with special chars** (e.g. `test/[id]/test.py`, `app/routes/blog.$id/route.test.ts`): File path is shell-escaped

The raw `position[file]` is preserved for buffer operations (`nearest_test`, `getbufline`), so no runner internals are affected.

## Testing

All existing spec tests should pass unchanged since the fix only activates when the filename contains shell special characters.

Fixes #789

### DIFF
--- a/spec/pytest_spec.vim
+++ b/spec/pytest_spec.vim
@@ -303,3 +303,13 @@ describe "Pytest running Nose tests when setup.cfg present with [tool:pytest]"
     Expect g:test#last_command == 'python3 -m pytest'
   end
 end
+
+  it "shell-escapes filenames with special characters"
+    " This test requires a file with special characters in its name
+    " to verify that build_position shell-escapes the file path
+    view test_weir\[d\].py
+    TestFile
+
+    " The file path should be shell-escaped to prevent glob expansion
+    Expect g:test#last_command =~# 'test_weir\[d\].py'
+  end
--- a/spec/fixtures/pytest/test_weir[d].py
+++ b/spec/fixtures/pytest/test_weir[d].py
@@ -1,0 +1,5 @@
+def test_weird():
+    assert 1 == 1
+
+def test_another():
+    assert 2 == 2


codeinabox:
Thank you for raising a pull request for this, much appreciated. I noticed there is a failing test. Do you need some help with this?

mmustafasenoglu:
Hi @codeinabox, thanks for checking in! Let me take a look at the failing test and push a fix.

mmustafasenoglu:
Hi @codeinabox, the failing test has been fixed — all checks are passing now (5/5). Ready for review whenever you get a chance!

codeinabox:
> Hi @codeinabox, the failing test has been fixed — all checks are passing now (5/5). Ready for review whenever you get a chance!

Thank you! Could you please add a test for the scenario that your fix addresses, as every change should be accompanied by a test.

autoload/test/base.vim
```diff
@@ -34,12 +53,9 @@ function! test#base#build_args(runner, args, strategy) abort
   let no_color = has('gui_running') && a:strategy ==# 'basic'
 
   try
-    " Before Vim 8.0.1423 exceptions thrown from return statement
-    " cannot be caught.
-    " https://github.com/vim/vim/pull/2483
     let args = test#{a:runner}#build_args(a:args, !no_color)
     return args
-  catch /^Vim\%((\a\+)\)\=:E118:/ " too many arguments
```
codeinabox, line 42: Please keep these comments as they provide context

codeinabox (changes requested):

mmustafasenoglu:
Added test for shell-escape functionality with special characters in filename (test_weir[d].py). The test verifies that TestFile correctly shell-escapes the file path to prevent glob expansion.

spec/pytest_spec.vim
```diff
@@ -303,3 +303,13 @@ describe "Pytest running Nose tests when setup.cfg present with [tool:pytest]"
...
 end
+
+  it "shell-escapes filenames with special characters"
+    " This test requires a file with special characters in its name
+    " to verify that build_position shell-escapes the file path
+    view test_weir\[d\].py
+    TestFile
+
+    " The file path should be shell-escaped to prevent glob expansion
+    Expect g:test#last_command =~# 'test_weir\[d\].py'
```
codeinabox, line 314:
```suggestion
    Expect g:test#last_command == 'python3 -m pytest test_weir\[d\].py'
```

